PR #85, opened 2022-06-28
Add link to v2 for unsupported name states

TateB:
- names cannot yet be registered, renewed, or imported in v3
- this PR adds a warning link to v2 for any state that isn't yet supported

### DIFF
--- a/src/hooks/useNameDetails.tsx
+++ b/src/hooks/useNameDetails.tsx
@@ -1,12 +1,14 @@
 import { useEns } from '@app/utils/EnsProvider'
 import { truncateFormat } from '@ensdomains/ensjs/dist/cjs/utils/format'
 import { ReactNode, useEffect, useMemo } from 'react'
+import { useTranslation } from 'react-i18next'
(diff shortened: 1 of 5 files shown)


src/hooks/useNameDetails.tsx
```diff
@@ -29,35 +28,73 @@ export const useNameDetails = (name: string) => {
...
+    }
+    if (
+      registrationStatus === 'available' ||
+      registrationStatus === 'premium' ||
+      registrationStatus === 'notImported' ||
+      registrationStatus === 'notOwned'
+    ) {
+      return (
+        <>
+          Some features are not yet available, but you can still{' '}
```
jefflau, line 59: i18n for this warning msg

jefflau (changes requested):
Out of scope of this PR, but can we get i18n for the error messages such as "This name is not valid."

jefflau (changes requested):
Can't this line just be `labels.length > 2`? Since it will never be a .eth name here

jefflau (approved):

PR #93, opened 2022-06-30
Gracefully display com.twitter when leading "@"

stoooops:
Love the new app!

# Issue
Users who write their `com.twitter` entry with leading "@" will see "@@" in the UI.

# Fix
Enhance display logic to ensure the "@" symbol always appears exactly once in the UI.

# Before
image

# After
image

# Author
Patched authored by cory.eth (Cory Gabrielsen) (@cory_eth).


LeonmanRolls:
Thanks very much @stoooops, glad you like the app!

PR #103, opened 2022-07-06
Add transaction flow

TateB:
Adds the transaction flow and transaction provider.


jefflau (changes requested):
Metamask pop ups before confirmation in ENS modal

Reproduction:

1) add a transaction
2) reject the transaction
3) close ENS modal
4) add same transaction again (metamask immediately pop ups instead of waiting for me to hit confirm in the ENS modal)

jefflau:
Screenshot 2022-07-07 at 16 19 27
Screenshot 2022-07-07 at 16 18 57

Also not super friendly when on smaller screens. Doesn't seem to take into account the height of my screen so I can't always see the title of the modal. And i can't scroll either

jefflau (approved):

PR #123, opened 2022-08-09
Dynamic load transaction input components

storywithoutend:


TateB (changes requested):
the dynamic import should have a loader specified so the user doesn't see a blank screen

TateB (approved):

PR #126, opened 2022-08-16
Friendly resolver

storywithoutend:


LeonmanRolls (changes requested):
Can you add a stateless test and also block the user from editing if they don't have a resolver set (so they can't click out of the overlay and get to the edit screen)

src/hooks/useProfile.test.ts
```diff
@@ -0,0 +1,21 @@
...
+const mockUseEns = mockFunction(useEns)
+const mockGetProfile = jest.fn()
+
+describe('useProfile', () => {
+  mockUseEns.mockReturnValue({
+    getProfile: mockGetProfile,
+    ready: true,
+  })
+
+  it('should call setProfile with the name', async () => {
```
TateB, line 16: getProfile?

src/hooks/useResolverStatus.ts
```diff
@@ -0,0 +1,136 @@
...
+type Result = {
+  status: {
+    hasLatestResolver: boolean
+    hasMigratedProfile: boolean
+    hasCreatedProfile: boolean
+  }
+  loading: boolean
+}
+
+export const useResolverStatus = (name: string, skip?: boolean, options?: Options): Result => {
```
TateB, line 56: can this be a useQuery function?

src/hooks/usePublicResolverAddress.ts
```diff
@@ -0,0 +1,22 @@
+import { useEns } from '@app/utils/EnsProvider'
+import { useQuery } from 'react-query'
+
+export const usePublicResolverAddress = () => {
```
TateB, line 4: you can make this synchronous if you use the getContractAddress function (just needs chainid), should probably add a generic useContractAddress hook but i'll do that separately

src/transaction-flow/input/ProfileEditor/ProfileEditor.tsx
```diff
@@ -459,26 +467,33 @@ const ProfileEditor = ({ data, dispatch, onDismiss }: Props) => {
         dispatch({ name: 'setFlowStage', payload: 'transaction' })
         return
       }
+
       dispatch({
         name: 'startFlow',
-        key: `update-profile-${name}`,
+        key: `edit-profile-flow-${name}`,
```
TateB, line 473: nit: flow is assumed so is doesn't really need to be added here

src/transaction-flow/input/ProfileEditor/ResolverWarningOverlay.tsx
```diff
@@ -0,0 +1,151 @@
...
+          name: 'TransferProfile',
+          data: { name },
+        },
+      },
+      key: `edit-profile-${name}`,
+    })
+  }
+
+  const title = (() => {
+    if (resumable) return 'Resume profile migration'
```
TateB, line 113: i18n, also why is this a func?

src/transaction-flow/transaction/migrateProfileWithSync.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  const contentHash = (() => {
+    if (data.records?.contentHash || data.records?.contentHash === '')
+      return contentHashToString(data.records.contentHash)
+    const afterCh = contentHashToString(profile.records.contentHash)
+    const beforeCh = contentHashToString(resolverProfile?.records?.contentHash)
+    if (beforeCh && !afterCh) return ''
+    if (afterCh) return afterCh
+    return undefined
+  })()
```
TateB, line 75: any reason this is a func and being called instead of just being inline?

TateB (changes requested):

src/hooks/usePublicResolverAddress.ts
TateB: useContractAddress:

```ts
import { useEns } from '@app/utils/EnsProvider'
import { ContractName } from '@ensdomains/ensjs/dist/cjs/contracts/types'
import { useChainId } from './useChainId'

export const useContractAddress = (contractName: ContractName) => {
  const chainId = useChainId()
  const { getContractAddress } = useEns()
  return getContractAddress(chainId as any)(contractName)
}
```

src/transaction-flow/input/ProfileEditor/ProfileEditor.tsx
storywithoutend: I use 'flow' to differentiate between multistep process that is resumable. Otherwise if I open an input with he key 'edit-profile-{name}', it will overwrite the progress on the transaction flow.

src/hooks/useProfile.test.ts
storywithoutend: edited

src/hooks/useResolverStatus.ts
storywithoutend: switched

src/hooks/usePublicResolverAddress.ts
storywithoutend: added

src/transaction-flow/input/ProfileEditor/ResolverWarningOverlay.tsx
storywithoutend: switched to dictionary

src/transaction-flow/transaction/migrateProfileWithSync.ts
storywithoutend: Over thought logic...

deploy/00_register_other.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  return true
+}
+
+func.id = 'register-other-names'
+func.tags = ['register-other-names']
+func.dependencies = ['LegacyETHRegistrarController']
+func.runAtTheEnd = true
+
+export default func
```
TateB, line 156: this is for normal legacy registrations as well, you don't need the other deploy file if youre using this one

e2e/specs/stateless/_profileEditor.spec.js
```diff
@@ -0,0 +1,70 @@
+import { acceptMetamaskAccess, connectFromExisting } from '../../setup'
+
+describe('Profile Editor', () => {
+  before(() => {
+    acceptMetamaskAccess(2)
+  })
+
+  it('should force a name on the old registry to ', () => {
```
TateB, line 8: to what?

TateB (changes requested):

deploy/00_register_other.ts
storywithoutend: I originally merged the two files and then I undid it with some name changes to try to be clearer.

The reason for the unmerge is that the runtime for 00_register_legacy is hard-coded in ensdomains/contracts and at that time the PublicResolver contract will not have been deployed and will throw an error. I think we can consider refactoring the tagging system and such so that contracts are all deployed first and then the data is added.

e2e/specs/stateless/_profileEditor.spec.js
storywithoutend: fixed

deploy/00_register_other.ts
TateB: oh right i see now, in this case you can just keep the normal registration and record setting in the original deploy file, then create a new deploy file that specifically sets records on the new resolver. doing that makes sure there's less duplication of code

LeonmanRolls (approved):

PR #180, opened 2022-10-12
Misc fixes

TateB:
- fixed no names message during names page load
- added hover states for buttons on names page
- removed weird looking expiry box on mobile
- fixed name from overflowing weirdly on profile page
- removed registration date when name is not 2ld .eth in more tab
- fix avatar crop usability issues


LeonmanRolls (changes requested):
'No names' still shows on first load 
Avatar upload zoom buttons don't work

PR #194, opened 2022-10-24
fix: misc bugs 2

TateB:
fixed:
- create subname transaction details order
- grace period name ownership fetching
- encoded subname links in subnames list
- names page expiry dates on wrapped names
- `/my/profile` redirects to `/profile/{name}`
- unknown name truncation in primary names list

new fixed:
- grace period names in names list
- removed lower error if transaction is pending/complete


LeonmanRolls (changes requested):
Please add appropriate tests to cover these fixes.

LeonmanRolls (approved):

PR #205, opened 2022-10-30
misc bugs 3

TateB:
fixed:
- dropdown visibility issues
- owner/manager label visibility
- empty names list
- header jumping on desktop


LeonmanRolls (changes requested):
Please add tests to prevent bug regression, probably can only do it for the names page one in this case

LeonmanRolls (approved):

Issue #114, opened 2022-08-02
The font used shows the wrong image for ➌ (U+278C)


cthulu-dot-eth:
The font used for the V3 manager app shows the wrong image for ➌ (U+278C) [[DINGBAT NEGATIVE CIRCLED SANS-SERIF DIGIT THREE](https://www.fontspace.com/unicode/char/278C-dingbat-negative-circled-sans-serif-digit-three)]
![Screenshot 2022-08-02 110610](https://user-images.githubusercontent.com/96479721/182337317-243ff6b6-2e56-47f0-a848-c9eb4dfef2fa.png)

And the ENS name shows up as invalid:
![invalid](https://user-images.githubusercontent.com/96479721/182338235-c1a44e14-af12-423f-9bfa-597e28a63b12.png)

It shows up correctly in the old manager app:
![Screenshot 2022-08-02 110713](https://user-images.githubusercontent.com/96479721/182337659-6162615f-18f8-4936-a222-089717773ec8.png)

But the metadata service image generation font is affected by the same problem:
https://github.com/ensdomains/ens-metadata-service/issues/119

ENS Name affected: ➌➌➌➌.eth [[V2 Manager app link](https://app.ens.domains/name/%E2%9E%8C%E2%9E%8C%E2%9E%8C%E2%9E%8C.eth/details)] [[V3 Manager app link](https://alpha.ens.domains/profile/%E2%9E%8C%E2%9E%8C%E2%9E%8C%E2%9E%8C.eth)]

PR #206, opened 2022-10-30
misc bugs 4

TateB:
fixed:
- wrong emoji on windows #114 
- header buttons behaviour on search focus


LeonmanRolls (changes requested):
Add test for the hidden header buttons

TateB:
> Add test for the hidden header buttons

@LeonmanRolls wasn't the concept to test for the bugs? if so, the header buttons being hidden was never the bug here

LeonmanRolls:
> > Add test for the hidden header buttons
> 
> 
> 
> @LeonmanRolls wasn't the concept to test for the bugs? if so, the header buttons being hidden was never the bug here

It came up as a bug because one would click the search bar and then the buttons would disappear, but the search bar didn't expand to take up the space.

TateB:
> It came up as a bug because one would click the search bar and then the buttons would disappear, but the search bar didn't expand to take up the space.

@LeonmanRolls wouldn't that mean that i want to test for the search bar expanding then?

LeonmanRolls:
> > It came up as a bug because one would click the search bar and then the buttons would disappear, but the search bar didn't expand to take up the space.
> 
> @LeonmanRolls wouldn't that mean that i want to test for the search bar expanding then?

Yes that's a more accurate way to say it

LeonmanRolls (approved):

PR #207, opened 2022-10-31
show errors relating to wrapper label byte limit

TateB:


src/components/pages/profile/[name]/Profile.tsx
```diff
@@ -78,15 +77,12 @@ const ProfileContent = ({ nameDetails, isSelf, isLoading, name }: Props) => {
...
+    canBeWrapped: _canBeWrapped,
   } = nameDetails
 
   const selfAbilities = useSelfAbilities(address, ownerData, name)
-  const nameWrapperExists = useWrapperExists()
   const canBeWrapped =
-    nameWrapperExists &&
-    !isWrapped &&
-    normalisedName.endsWith('.eth') &&
+    _canBeWrapped &&
```
Comment (LeonmanRolls) [inline src/components/pages/profile/[name]/Profile.tsx:85]: shouldn't the 'private' variable be the one that is only used in this component? not the one coming from nameDetails

LeonmanRolls (approved):

PR #210, opened 2022-10-31
add Intl.Segmenter polyfill for firefox support

mdtanrikulu:


src/components/@molecules/NFTTemplate/NFTTemplate.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+        backgroundImage={undefined}
+        isNormalised
+      />,
+    )
+    expect(getByText('thisnameislonge')).toBeInTheDocument()
+    expect(getByText('rthan25char.eth')).toBeInTheDocument()
+  })
+
+  it('should use polyfill of Intl.Segmenter if browser does not support', async () => {
+    ;(window.Intl.Segmenter as typeof Intl['Segmenter']) = undefined as any // override native api
```
storywithoutend, line 52: remove comment

src/components/@molecules/NFTTemplate/NFTTemplate.tsx
```diff
@@ -54,11 +54,12 @@ const Text = styled.text(
 
 const NFTTemplate = ({ name, backgroundImage, isNormalised }: Props) => {
   const Satoshi = new FontFace('Satoshi', 'url(/fonts/sans-serif/Satoshi-Bold.otf)')
-  const DejaVuSans = new FontFace('DejaVuSans', 'url(/fonts/sans-serif/DejaVuSans-Bold.ttf)')
```
storywithoutend, line 57: Is this here to make sure that the font face is loaded before making "elementData" calculations? If so, then they should probable be wrapped in a useEffect similar to loading the polyfill

src/components/@molecules/NFTTemplate/NFTTemplate.tsx
```diff
@@ -70,7 +71,7 @@ const NFTTemplate = ({ name, backgroundImage, isNormalised }: Props) => {
     if (isSubdomain && !name.includes('...')) {
       subdomain = `${labels.slice(0, labels.length - 2).join('.')}.`
       domain = labels.slice(-2).join('.')
-      if (getSegmentLength(subdomain) > MAX_CHAR) {
+      if (getSegmentLength?.(subdomain) > MAX_CHAR) {
```
storywithoutend, line 74: The "?." is unnecessary since getSegmentLength will always return a value. I added this when I was doing a quick test and had a different structure.

storywithoutend (changes requested):
The testing structure has changed quiet a bit since when you created this branch. Please resolve conflicts and merge dev branch into your branch to test again.

src/components/@molecules/NFTTemplate/NFTTemplate.tsx
mdtanrikulu: I think we can just move font load outside of the the component, since it has no dependency or bound with react, will be a constant.

mdtanrikulu: Based on chat discussion, we do not need to re-load the Satoshi font, since it is preloaded in the main style of the app.

storywithoutend (approved):

PR #218, opened 2022-11-10
Dogfood

LeonmanRolls:


storywithoutend (changes requested):
Only issue I found was that the latest resolver 0x84eA74d481Ee0A5332c457a4d796187F6Ba67fEB does not work and breaks the ability to edit records.

Other than this. LGTM

PR #217, opened 2022-11-08
[DO NOT MERGE] Web3modal int

makoto:
This is a fork of https://github.com/ensdomains/ens-app-v3/pull/144 to run deploy preview

PR #220, opened 2022-11-14
feat: replace with new w3m version

crypblizz8:
Overview
- Iteration of #144 & #217 
- Replaced RainbowKit connectModal
- No need to replace wagmi hooks


src/hooks/useBasicName.ts
```diff
@@ -87,6 +87,6 @@ export const useBasicName = (name?: string | null, normalised?: boolean) => {
       !isWrapped &&
       normalisedName?.endsWith('.eth') &&
       !isLabelTooLong(normalisedName),
-    isCachedData: status === 'success' && isFetched && !isFetchedAfterMount,
```
LeonmanRolls, line 90: Why has !isFetchedAftterMount been removed?

LeonmanRolls (changes requested):

PR #226, opened 2022-11-21
fix flaky position on avatar load

mdtanrikulu:
When user picks avatar image with file picker multiple times, the loaded image appears in wrong position and size.
This PR avoid the side effect by debouncing the load callback function.

flaky-avatar-load

### DIFF
--- a/src/components/@molecules/ProfileEditor/Avatar/AvatarCrop.tsx
+++ b/src/components/@molecules/ProfileEditor/Avatar/AvatarCrop.tsx
@@ -1,4 +1,5 @@
 /* eslint-disable no-multi-assign */
+import { debounce } from 'lodash'
 import { useCallback, useEffect, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import styled, { css } from 'styled-components'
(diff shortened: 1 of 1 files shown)


TateB:
there's a `useDebouncedCallback` hook that i think you could use here instead of lodash

TateB (changes requested):
see comment

mdtanrikulu:
> `useDebouncedCallback`

I haven't see dependency option in that hook, how do we manage the following pattern? Or should I just extend that capability in hook itself?

```js
useCallback(debounce(() => {}, 100), [dependencyFn])
```

```js
useDebouncedCallback(() => {}, 100)
```

TateB:
could change the hook to this i think:

```js
export default function useDebouncedCallback<T extends (...args: any[]) => ReturnType<T>>(
  func: T,
  wait?: number,
  deps: DependencyList = [],
): T {
  const timerId = useRef<NodeJS.Timer>()

  return useCallback(
    (...args: Parameters<T>) => {
      clearTimeout(timerId.current)
      timerId.current = setTimeout(() => func(...args), wait)
    },
    [func, wait, ...deps],
  ) as T
}
```

TateB (approved):

PR #228, opened 2022-11-21
fix flashing error message on send name modal

storywithoutend:


src/components/@molecules/TransactionDialogManager/TransactionDialogManager.tsx
```diff
@@ -101,7 +101,7 @@ export const TransactionDialogManager = ({
   }, [selectedKey, selectedItem, onDismiss, dispatch, t])
 
   return (
-    <Dialog variant="blank" open={!!state.selectedKey} onDismiss={onBackgroundDismiss}>
+    <Dialog variant="blank" open={!!selectedKey} onDismiss={onBackgroundDismiss}>
```
TateB, line 104: pretty sure it was previously `state.selectedKey` to make sure that that dismiss animation is finished before unloading the modal

src/hooks/useBasicName.ts
```diff
@@ -25,8 +25,8 @@ export const useBasicName = (name?: string | null, normalised?: boolean) => {
     status,
   } = useQuery(
     ['batch', 'getOwner', 'getExpiry', normalisedName],
-    () => {
-      const batchQueries = addRegistrationStatusToBatch(ens, normalisedName)
+    async () => {
```
TateB, line 28: this func doesn't need to be async since a promise is returned either way

src/hooks/useBasicName.ts
```diff
@@ -25,8 +25,8 @@ export const useBasicName = (name?: string | null, normalised?: boolean) => {
...
+      const batchQueries = await addRegistrationStatusToBatch(ens, normalisedName)
```
TateB, line 29: this func isn't async, await won't do anything

TateB (changes requested):

LeonmanRolls (approved):

PR #235, opened 2022-11-23
fix: batch types in useBasicName

TateB:


storywithoutend (approved):

storywithoutend:
Code looks good. Can you fix tests and merge? Looks like issue is from deleting useRegistrationStatus. I think the value of registrationStatus in useBasicName has a bug, I can go in and fix that after.

PR #278, opened 2023-01-24
reimplement profile editor record input

storywithoutend:


src/utils/validate.ts
```diff
@@ -1,3 +1,5 @@
+import { getAddress } from 'ethers/lib/utils'
```
TateB, line 1:
```suggestion
import { getAddress } from '@ethersproject/address'
```

use `@ethersproject/*` packages instead of ethers

src/utils/validate.ts
```diff
@@ -15,8 +17,14 @@ export const validateCryptoAddress =
   (coin: string) =>
   async (address: string): Promise<string | boolean> => {
```
TateB, line 18: why was this async? promises in validation are usually not great

TateB (changes requested):

src/utils/validate.ts
TateB: only noticed because you switched the references to this func to use `await`

storywithoutend: The validateCrypto function is async, I checked inside and it doesn't have an async function inside so I turned it into a sync func.

TateB (approved):
add pointer cursor to record delete button, then lgtm

PR #279, opened 2023-01-25
fix: allow unknown label wrapping

TateB:
- adds ability to wrap a name with an unknown label
- a dialog will pop up asking for the correct labels for the name
- also adds enhanced label decryption by fetching the name from the namewrapper, as well as fetching from the subgraph to attempt to match the name/labels. if a match is found, a user on the profile page will the immediately redirected to the corrected page with the decrypted label. the label will also be saved to storage.
- depends on https://github.com/ensdomains/ensjs-v3/pull/103

todo:
- [x] cypress stateless test for wrapping unknown label
- [x] cypress stateful test for label fetching
- [x] unit test updates


src/components/pages/profile/[name]/Profile.tsx
```diff
@@ -150,6 +145,31 @@ const ProfileContent = ({ nameDetails, isSelf, isLoading, name }: Props) => {
...
 
+  useEffect(() => {
+    if (
+      name !== profile?.decryptedName &&
+      profile?.decryptedName &&
+      !isSelf &&
+      getEncryptedLabelAmount(normalisedName) > getEncryptedLabelAmount(profile.decryptedName)
+    ) {
+      router.replace(`/profile/${profile.decryptedName}`, { shallow: true, maintainHistory: true })
+    } else if (
```
Comment (LeonmanRolls) [inline src/components/pages/profile/[name]/Profile.tsx:156]: Can you move this into its own named hook or add a brief comment on what it's doing

src/hooks/useValidate.ts
```diff
@@ -10,9 +16,18 @@ export const useValidate = (input: string, skip?: any) => {
     let _valid: boolean | undefined
     if (!skip) {
       try {
-        _normalisedName = validateName(decodeURIComponent(input))
+        let _input = decodeURIComponent(input)
```
LeonmanRolls, line 19: imo '_' should be used as a last resort, this could be called 'decodedInput'

LeonmanRolls (approved):

PR #298, opened 2023-02-08
feat: add correct wrapped subname tags + expiry date

TateB:
fixes/features:
- [x] support for displaying negative expiry on subname
  - negative expiry only applies to subnames with PCC burned
  - to test: navigate wrapped-expired-subnames.eth and then click the subnames tab, the "sub" subname should have a negative expiry
- [x] show subnames on names page
- [x] fix ownership tags always showing as enabled (+ more tests to prevent regression)
- [x] show expiry in /my/names list for unwrapped .eth names where user is controller but not registrant
- [x] show expiry of PCC subnames
- [x] show PCC expired subnames in subnames list
- [x] style tweak: make label for names in /my/names black, rest of name grey (previously only tld was grey)
- [x] show "not owned" when pcc expired in subnames list
- [x] show "not owned" and correct expiry on PCC expired subname pages


LeonmanRolls (changes requested):
Remove yalc

LeonmanRolls (dismissed):

LeonmanRolls (approved):

PR #299, opened 2023-02-08
fix: misc bugs 5

TateB:
fixed:
- ipad/safari footer is not aligned with bottom of page (stupid workaround, i shouldn't need to do this)
- extra horizontal line on profile card after disconnecting (mobile)
- ios/safari search input focus on click with keyboard
- "Not Supported" flash in search for DNS names
- use "Imported" for dns names in search instead of "Registered"
- use "Owned" for subnames in search instead of "Registered"
- avatars reloading too frequently
- avatar uploader scrollbars/buggy/non-centred images

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -26,7 +26,7 @@ const customJestConfig = {
     '@ensdomains/ensjs/(.*)$': '@ensdomains/ensjs/dist/cjs/$1',
   },
   setupFilesAfterEnv: ['<rootDir>/jest/jest.setup.ts'],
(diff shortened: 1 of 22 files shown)


jefflau (approved):

jefflau:
Avatar preview shows cropped image. But when you go back to the records view, it's uncropped. When you complete registration, the image is also cropped fine after.

Also - should NFT image at the end of registration show the NFT avatar? It currently just shows up as the default blue.

PR #311, opened 2023-02-15
add search query state to url history

storywithoutend:


src/components/pages/my/names/MyNames.tsx
```diff
@@ -69,9 +69,53 @@ const MyNames = () => {
...
+  /* eslint-disable no-nested-ternary */
+  const [sortType, setSortType] = useState<SortType | undefined>(
+    sort === 'expiry'
+      ? SortType.expiryDate
+      : sort === 'name'
+      ? SortType.labelName
+      : sort === 'creation'
+      ? SortType.creationDate
+      : undefined,
+  )
```
TateB, line 88: why not just use either enums directly, or just a string

src/components/pages/my/names/MyNames.tsx
```diff
@@ -69,9 +69,53 @@ const MyNames = () => {
...
+  }, [sortDirection])
+
+  const [searchQuery, setSearchQuery] = useState(search || '')
+  useEffect(() => {
+    const url = new URL(router.asPath, window.location.origin)
+    if (searchQuery) url.searchParams.set('search', searchQuery)
+    else url.searchParams.delete('search')
+    router.replace(url.toString(), undefined, { shallow: true })
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [searchQuery])
```
TateB, line 118: do you think you could create an abstraction of these hooks? looks like you're using very similar logic everywhere

TateB (changes requested):

src/components/pages/my/names/MyNames.tsx
```diff
@@ -69,9 +71,13 @@ const MyNames = () => {
       setSelectedNames([...selectedNames, name])
     }
   }
-  const [sortType, setSortType] = useState<SortType | undefined>()
-  const [sortDirection, setSortDirection] = useState<SortDirection>(SortDirection.desc)
-  const [searchQuery, setSearchQuery] = useState('')
+
+  const [sortType, setSortType] = useQueryParameterState<SortType>('sort', 'expiryDate' as SortType)
```
TateB, line 75: if you use e.g.
```ts
type SortType = 'expiryDate' | 'labelName' | 'creationDate'
```
you don't need to typecast

TateB (approved):

PR #317, opened 2023-02-20
fix: misc bugs 6

TateB:
fixed:
- FET-955: searching for emojis hangs profile page (not locally, only on cloudflare)
  - test: cypress/stateful/profile
- FET-957: background for transparent avatar uploads is black instead of white
  - test: N/A for now, full testing for avatar uploads to be implemented later
- FET-945: continue button in transaction toasts
  - test: cypress/stateless/registerName
- FET-967: primary name flow resumable even if only 1 tx
  - test: cypress/stateless/setPrimary
- FET-961: create subname + unknown labels flow allows dot in middle of input
  - test: unit/useValidateSubnameLabel
- FET-953: token section in more tab is shown for any .eth subname
  - added: ownership section is always shown in more tab, even if name is not owned
  - test: unit/MoreTab
- FET-935: "I'd like to setup my profile first" button in registration is black and regular weight instead of blue/bold
  - test: cypress/stateless/reigsterName
- FET-943: PCC expired subnames show "delete subname" button but not actually possible
  - test: cypress/stateless/deleteSubname
- FET-959: two close icons in "Add profile fields" dialog
  - test: N/A, can't really test for an unexpected scenario like this
- FET-958: Add profile fields dialog search does not behave correctly, e.g. searching for website does not yield website record result nor contenthash results
  - test: unit/AddProfileRecordView

PR #319, opened 2023-02-22
fix: misc bugs 7

TateB:
relies on #317 
fixes:
- FET-952: names with encrypted/unknown labels not properly set as primary name
  - notes:  made the `UnknownLabels` input item fully generic, so it can be used by the name wrapping flow as well as the primary name flow.
  - test: cypress/stateless/setPrimary
 - FET-950: expiry for emancipated subname is not shown in profile's more tab
   - notes: added `pccExpired` to the `useBasicName()` data so does not need to be calculated multiple times; renamed `useNameDates()` to `useRegistrationData()` and removed the query for `expiryDate` query within it, meaning expiryDate displayed in more tab is always from onchain data.
   - test: unit/MoreTab + unit/Miscellaneous
 - FET-939: etherscan link in edit resolver dialog is incorrect
   - test: cypress/stateless/updateResolver
 - FET-938: view link in more tab for 2ld .eth doesn't link to registration transaction
   - test: unit/Miscellaneous
 - FET-937: when changing primary name and only the currently set primary name is eligible, dialog looks broken
   - notes: also fixed avatars loading in the primary names list when the avatar has an encrypted label
   - test: unit/SelectPrimaryName
 - FET-932: dns import page does not have tab title
   - test: N/A
 - FET-930: items within the registrar dropdown on dns import page don't take up the full width of the dropdown
   - test: N/A, added comments
 - FET-927: back button on dns import page is not correct size
   - notes: this issue exists because DNS import doesn't use the standard page layout. will need to change in future
   - test: N/A
 - FET-929: translation keys on dnssec import page will sometimes show
   - test: N/A for now, could add check for bundled i18n files in future


LeonmanRolls (changes requested):
Add tests please

LeonmanRolls (approved):

PR #325, opened 2023-02-24
Disable editing records when resolver is not namewrapper aware

LeonmanRolls:


src/transaction-flow/input/ProfileEditor/ProfileEditor-flow.tsx
```diff
@@ -192,6 +198,14 @@ const ProfileEditor = ({ data = {}, transactions = [], dispatch, onDismiss }: Pr
     }
   }, [status, statusLoading, resumable])
 
+  const { isWrapped } = useNameDetails(name)
```
TateB, line 201: since you don't need all of the data from `useNameDetails()` would be better to just take the calculation for `isWrapped` out, make it a util func, and make a new hook called `useOwner()` or similar

src/utils/utils.spec.ts
```diff
@@ -42,3 +49,18 @@ describe('formatExpiry', () => {
     expect(result).toEqual('\nJanuary 1, 2020')
   })
 })
+
+describe('canEdirRecordsWhenWrappedCalc', () => {
```
TateB, line 53: nit: spelling

src/components/pages/profile/[name]/tabs/RecordsTab.tsx
```diff
@@ -168,6 +174,17 @@ export const RecordsTab = ({
       { name },
       { disableBackgroundClick: true },
     )
+
+  const { profile } = useProfile(name, name !== '')
```
Comment (TateB) [inline src/components/pages/profile/[name]/tabs/RecordsTab.tsx:178]: profile doesn't need to be fetched here

src/components/@molecules/DisabledButtonWithTooltip.tsx
```diff
@@ -26,8 +18,8 @@ export const DisabledButtonWithTooltip = ({
   mobilePlacement = 'left',
   width = 250,
   mobileWidth = 150,
-  buttonWidth,
-  mobileButtonWidth,
+  buttonWidth = undefined,
```
TateB, line 21: this change isn't needed

TateB (changes requested):

src/transaction-flow/input/ProfileEditor/ProfileEditor-flow.tsx
LeonmanRolls: isn't useQuery caching anything heavy

TateB (approved):

PR #331, opened 2023-02-28
Fet956 expired name incorrect details

LeonmanRolls:


src/hooks/useNameDetails.tsx
```diff
@@ -57,18 +59,37 @@ export const useNameDetails = (name: string) => {
       return t('errors.invalidName')
     }
     if (registrationStatus === 'gracePeriod') {
-      return t('errors.expiringSoon')
+      return `${t('errors.expiringSoonOne')} ${gracePeriodEndDate?.toString()} ${t(
+        'errors.expiringSoonTwo',
+      )}`
```
TateB, line 64: the formatting on this could change depending on the translation, should use translation variable in a single translation

src/hooks/useNameDetails.tsx
```diff
@@ -57,18 +59,37 @@ export const useNameDetails = (name: string) => {
...
+    ready,
+    registrationStatus,
+    status,
+    t,
+    valid,
+  ])
+
+  const errorTitle = useMemo(() => {
+    if (registrationStatus === 'gracePeriod') {
+      return `${name}  has expired`
```
TateB, line 84: translation

TateB (changes requested):
does this solve the problem outlined in the ticket? there aren't many details in it but as far as i can tell the solution is to not show the "owner" value when it's expired

PR #332, opened 2023-03-01
fix: continue button in notification throws error when resumable transactions are completed

storywithoutend:
I looked into updating the notification but that would require a lot of work. The next best thing was to add a guard to do nothing if the transaction item no longer exists.


LeonmanRolls (changes requested):
Unit tests please!

PR #349, opened 2023-03-08
add input to plus minus control

storywithoutend:


TateB (changes requested):
- show the suffix ("years") all the time
- buttons shouldn't be disabled when input is focused (clicking the buttons defocuses the input anyway so they still work)
- should probably highlight the whole input value when focused so that they can just immediately enter a new one. if they are entering a manual input they probably want to enter something completely different anyway, rather than adding to what is there

PR #352, opened 2023-03-09
fix: search history order

TateB:
to test: search for some names/addresses and click on them

search history should be in last searched order now (recent first)


LeonmanRolls (changes requested):
Tests please

LeonmanRolls (approved):

PR #368, opened 2023-03-15
Fix transactions section in settings

storywithoutend:
Latest Update
* Updated the "No primary name set" to match the "No recent transactions" style

Update Outlink red styling
* Go to settings
* Create a failed transaction
* Etherscan link should be red

Transaction List items
* If there is no continue button, do not render container div
* Add padding and remove set height restrictions


TateB:
FYI removing the height values also removes the height transitions (auto height doesn't transition)

TateB (approved):

PR #371, opened 2023-03-17
Plus minus control with input - No decimal support

storywithoutend:
Change Requests
* show the suffix years - Will be addressed when Dom rethinks the design to handle setting specific date or specific unit.
* button shouldn't be defocused when input is focused - Update control so button content is not clickable when disabled. Can discuss further if buttons should be disabled or not.
* should probably highlight whole input when focused - Added select value on focus


TateB (changes requested):
can you add `inputmode="numeric"` and `pattern="[0-9]*" to the input for mobile number inputs

TateB (approved):

PR #381, opened 2023-03-28
HTTP Headers

LeonmanRolls:


package.json
```diff
@@ -20,7 +20,7 @@
     "lint": "next lint && pnpm stylelint stylelint \"./src/**/*.tsx\"",
     "lint:types": "tsc --noEmit",
     "lint:fix": "next lint --fix",
-    "export": "next export",
+    "export": "next export && cp _headers ./out/_headers",
```
TateB, line 23: if you move `_headers` to `/public` you don't need to manually copy over

TateB (changes requested):

TateB (changes requested):

PR #474, opened 2023-04-25
add 307 redirects for v2 app links

gskril:
maintain SEO and backward link compatibility

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -79,6 +79,30 @@ let nextConfig = {
       },
     ]
   },
+  async redirects() {
+    return [
+      {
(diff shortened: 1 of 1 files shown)


LeonmanRolls (changes requested):
Thanks, I don't think this works, see: https://v2-redirects.ens-app-v3.pages.dev/name/unwrapped002.eth

I think the cloudflare rules are fine? Alternatively can define redirects in the _middleware.js file

gskril:
Weird, it's working locally but not on the cloudflare branch preview. I guess the cloudflare rules are sufficient

PR #540, opened 2023-07-06
Merge useSelfAbilities + useSubnameAbilities & Allow wrapped parent to delete unwrapped child and vise versa

storywithoutend:
* Merge useSelfAbilities and useSubnameAbilities into useAbilities
  * Update conditionals to use ts-pattern
  * Update canEdit so that it only factors if user has editing abilities (nothing to do with resolver)
  * Added canEditRecords to factor in if resolver is authorized or not.
  * Ported tests from original hooks over
  * Update canEdit usage to reflect changes
* Merged in PR to allow parent to delete subname in the following conditions
  * wrapped parent - unwrapped child
  * unwrapped parent - wrapped child


src/hooks/abilities/useAbilities.ts
```diff
@@ -0,0 +1,74 @@
...
+        basicNameData,
+        parentBasicNameData,
+        hasSubnames: hasSubnamesData.hasSubnames!,
+        t,
+      }),
+      ...getReclaimAbilities({
+        address,
+        basicNameData,
+        parentBasicNameData,
+      }),
```
TateB, line 56: can you add an explicit `UserAbilities` type or something similar?

src/hooks/abilities/useAbilities.ts
```diff
@@ -0,0 +1,74 @@
...
+  const parent = name?.split('.')?.slice(1).join('.')
+
+  const { address } = useAccountSafely()
+  const basicNameData = useBasicName(name, { skipGraph: false })
+  const resolverAuthorisation = useResolverIsAuthorized(name, {
+    enabled: !!name,
+  })
+  const parentBasicNameData = useBasicName(parent, { skipGraph: false, enabled: !!parent })
+
+  const hasSubnamesData = useHasSubnames(name)
```
TateB, line 26: is this hook loaded at any point when the user isn't connected, if so these should all be conditionally enabled by it

src/hooks/abilities/utils/getDeleteAbilities.ts
```diff
@@ -0,0 +1,185 @@
...
+  hasSubnames: boolean
+  t: TFunction
+}): Result => {
+  if (!checkSubname(name)) return BASE_ABILITIES
+  return (
+    match([basicNameData, parentBasicNameData])
+      /* --------------- WRAPPED WITH PCC BURNED --------------- */
+      /*
+       * WRAPPED NAME OWNER
+       * */
```
TateB, line 44: nit - maybe make these comments more friendly

src/hooks/abilities/utils/getDeleteAbilities.ts
```diff
@@ -0,0 +1,185 @@
...
+          },
+          {
+            ownerData: {
+              owner: P.select('parentOwner'),
+            },
+          },
+        ],
+        ({ fuses, parentOwner }) => ({
+          ...BASE_ABILITIES,
+          canDelete: !hasSubnames && !fuses?.CANNOT_TRANSFER,
```
TateB, line 67: fuses object can be forcefully unwrapped here, it will always be defined

src/hooks/abilities/utils/getDeleteAbilities.ts
```diff
@@ -0,0 +1,185 @@
...
+            },
+          },
+          {
+            ownerData: {
+              owner: P.select('parentOwner'),
+            },
+          },
+        ],
+        ({ fuses, parentOwner }) => ({
+          ...BASE_ABILITIES,
```
TateB, line 66: what's the point in doing this? you overwrite the canDelete property in the line after

src/hooks/abilities/utils/getDeleteAbilities.ts
```diff
@@ -0,0 +1,185 @@
...
+      .with(
+        [
+          {
+            ownerData: {
+              ownershipLevel: 'nameWrapper',
+              owner: P.when((owner) => owner === address),
+            },
+            wrapperData: {
+              parent: {
+                PARENT_CANNOT_CONTROL: P.not(true),
```
TateB, line 87: is this value ever going to be undefined if the parent object is defined?

src/hooks/abilities/utils/getDeleteAbilities.ts
```diff
@@ -0,0 +1,185 @@
...
+              },
+            },
+          },
+          {
+            ownerData: {
+              ownershipLevel: P.select('parentOwnershipLevel'),
```
TateB, line 93: can remove since it isn't used

src/hooks/abilities/utils/getEditAbilities.ts
```diff
@@ -0,0 +1,66 @@
...
+          canCreateSubdomains: !fuses.CANNOT_CREATE_SUBDOMAIN,
+          canEditTTL: !fuses.CANNOT_SET_TTL,
+        }
+      },
+    )
+    .with(
+      [
+        {
+          ownerData: {
+            ownershipLevel: P.not('nameWrapper'),
```
TateB, line 44: this can be removed right? it's already implied

src/hooks/abilities/utils/getReclaimAbilities.test.ts
TateB, line 1: kinda seems like these tests are more in-depth than they need to be

src/hooks/abilities/utils/getSendAbilities.ts
```diff
@@ -0,0 +1,422 @@
...
+  return (
+    match(basicNameData)
+      /*
+       * WRAPPED NAME OWNER OR MANAGER
+       * */
+      .with(
+        {
+          ownerData: { ownershipLevel: 'nameWrapper', owner: P.when((owner) => owner === address) },
+          wrapperData: {
+            parent: { PARENT_CANNOT_CONTROL: P.select('isOwner') },
```
TateB, line 50: since this is 2ld eth, this will always be true if the name is wrapped

src/hooks/abilities/utils/getSendAbilities.ts
```diff
@@ -0,0 +1,422 @@
...
+          },
+        }),
+      )
+      /* *
+       * UNWRAPPED NAME MANAGER
+       * */
+      .with(
+        {
+          ownerData: {
+            ownershipLevel: P.not('nameWrapper'),
```
TateB, line 112: this is already implied

src/hooks/abilities/useAbilities.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  const isLoading =
+    basicNameData.isLoading || parentBasicNameData.isLoading || hasSubnamesData.isLoading
+
+  const isCachedData = basicNameData.isCachedData || hasSubnamesData.isCachedData
+
+  const data = useMemo(() => {
+    if (!name) return undefined
+    if (isLoading) return undefined
```
TateB, line 35: when referencing abilities there is a lot of optional chaining with a default value, e.g. `abilities?.canSend ?? false`, maybe it makes sense to provide a default abilities value?

TateB (changes requested):

src/hooks/abilities/utils/getEditAbilities.ts
storywithoutend: I personally would prefer to leave it there so it makes the declarative case more clear.

LeonmanRolls (approved):

PR #559, opened 2023-08-07
Further safe fixes

LeonmanRolls:
Apply #552 to transaction refactor

Testing:
- try extending a name via safe app


package.json
```diff
@@ -148,6 +148,7 @@
     "@typescript-eslint/eslint-plugin": "^5.35.1",
     "@typescript-eslint/parser": "^5.35.1",
     "@wagmi/core-cjs": "npm:@wagmi/core@0.10.11-cjs",
+    "@web3modal/standalone": "^2.4.3",
```
TateB, line 151: where is this being used?

package.json
```diff
@@ -179,6 +180,7 @@
     "jest-canvas-mock": "^2.4.0",
     "jest-environment-jsdom": "^27.5.1",
     "jest-localstorage-mock": "^2.4.22",
+    "lokijs": "^1.5.12",
```
TateB, line 183: where is this being used?

LeonmanRolls: both being used in a unit test, because i couldn't get the instanceof check to pass in useIsSafe, there might be a better way though

LeonmanRolls: ok wasn't that hard to fix, can remove these too

PR #568, opened 2023-08-09
Timezone Record Parsing

lucemans:
This PR adds the ability to show the profile's **timezone** aswell as show the viewer the difference to their localtime
image
image


lucemans:
Might change this to show distance to UTC, as the `+01:00` syntax tends to be reserved for "offsets from UTC"
Or change it to something like "2 hours ahead", so it is more clear.

Thoughts?

PR #569, opened 2023-08-14
fix: extend warning not showing when user is owner but not manager

TateB:
previously, in certain situations when extending a name, it could show the ownership warning even when the user was the owner.

changes:
- added check function for showing warning
  - checks edit OR send ability
- use check function in profile snippet + more tab

Fixes FET-1268


src/utils/abilities/shouldShowExtendWarning.ts
```diff
@@ -0,0 +1,6 @@
+import { Abilities } from '@app/hooks/abilities/useAbilities'
+
+export const shouldShowExtendWarning = (abilities: Abilities | undefined) => {
```
LeonmanRolls, line 3: Nice helper function

PR #612, opened 2024-01-09
feat: subgraph latency check and error handling

talentlessguy:
- [x] latency reporting
- [x] network error reporting
- [x] recover from errors if request is successful again
- Pages
  - [x] `:name`
  - [x] `more`
  - [x] set primary name


talentlessguy:
image

to test:

- Open Chrome DevTools -> Create custom network rule
- Set latency to 5000ms
- Apply rule and reload the page

talentlessguy:
image

the hook works for failed requests too

i also narrowed down the hook to only subgraph errors

src/components/pages/profile/[name]/tabs/MoreTab/MoreTab.tsx
```diff
@@ -42,8 +45,8 @@ const MoreTab = ({ name, nameDetails, abilities }: Props) => {
       )}
       <Resolver
         name={name}
-        canEdit={abilities.canEdit}
-        canEditResolver={abilities.canEditResolver}
+        canEdit={!hasGlobalError && abilities.canEdit}
+        canEditResolver={!hasGlobalError && abilities.canEditResolver}
```
Comment (TateB) [inline src/components/pages/profile/[name]/tabs/MoreTab/MoreTab.tsx:49]: afaik this isn't parity behaviour, not sure that this behaviour is ideal because the error could be unrelated

src/hooks/errors/useHasGlobalError.ts
```diff
@@ -1,18 +1,18 @@
...
-    return (
-      activeHashes.filter((hash) => {
-        if (includeLatency) return !!errors[hash]
-        return !!errors[hash] && errors[hash].type !== 'ENSJSNetworkLatencyError'
-      }).length > 0
-    )
-  }, [globalState, includeLatency])
+  if (includeLatency || subgraphDebug === 'ENSJSSubgraphLatency') {
+    return Boolean(error || slow)
+  }
```
TateB, line 11: should this trigger when includeLatency is false but subgraphDebug is set to latency? also if subgraphDebug is latency and we want to trigger it, i assume its meant to be `true`.
i.e.
```suggestion
  if (includeLatency) {
    if (subgraphDebug === 'ENSJSSubgraphLatency') return true
    return Boolean(error || slow)
  }
```

src/hooks/useLocalStorage.ts
```diff
@@ -50,6 +52,37 @@ export const useLocalStorage = <D>(
   return [value, setValue]
 }
 
+type Value<T> = T | null
```
TateB, line 55: nit: more descriptive type name

src/hooks/useLocalStorage.ts
```diff
@@ -7,9 +8,10 @@ const isBrowser = !!(
   window.document.createElement
 )
 
-const getStorageValue = <D>(key: string, defaultValue: D) => {
+const getStorageValue = <D>(key: string, defaultValue: D): D => {
```
TateB, line 11: this won't always return type `D`

src/hooks/useLocalStorage.ts
```diff
@@ -50,6 +52,37 @@ export const useLocalStorage = <D>(
...
+    (event: StorageEvent | CustomEvent) => {
+      if ((event as StorageEvent)?.key && (event as StorageEvent).key !== key) {
+        return
+      }
+      setStoredValue(readValue())
+    },
+    [key, readValue],
+  )
+
+  useEvent('storage', handleStorageChange)
```
TateB, line 82: if you're using the `useEvent` hook here, it's probably worth propagating it to `useLocalStorage` as well

src/transaction-flow/input/SelectPrimaryName/SelectPrimaryName-flow.tsx
```diff
@@ -395,7 +398,7 @@ const SelectPrimaryName = ({ data: { address }, dispatch, onDismiss }: Props) =>
           <Button
             data-testid="primary-next"
             onClick={onConfirm}
-            disabled={!selectedName || isLoadingName}
+            disabled={!selectedName || isLoadingName || hasGlobalError}
```
TateB, line 401: non-parity

TateB (changes requested):
minor issues, also add unit tests for `useHasSubgraphSyncErrors`

src/hooks/useLocalStorage.ts
TateB: actually nvm this should be fine

src/components/pages/profile/[name]/tabs/MoreTab/MoreTab.tsx
Comment (talentlessguy) [reply src/components/pages/profile/[name]/tabs/MoreTab/MoreTab.tsx:49]: depends on whether any of the nameDetails are fetched via a subgraph. or actually it doesn't matter - it just editrs a re

src/hooks/useLocalStorage.ts
talentlessguy: changed to `LocalStorageValue`

src/utils/useHasSubgraphSyncErrors.test.ts
talentlessguy, line 1: should I include more test cases or just two is fine?

@TateB

TateB: this is fine

TateB (approved):

PR #626, opened 2024-02-02
Offchain claim process

talentlessguy:
this PR adds off-chain claim process flow.

Here are the steps: DNSSec verification, offchain ownership verification, claim finished

Screenshots:

image

image


talentlessguy:
there's no check for the resolver in the designs so i'm not sure how to handle that. should we or should we not add a warning / error in case of mismatches in resolver addresses?

src/hooks/dns/useDnsOffchainStatus.ts
```diff
@@ -54,28 +53,19 @@ export const useDnsOffchainStatus = ({ name, enabled = true }: UseDnsOffchainSta
...
+  const isCachedData = isDnsOffchainDataCachedData
 
   const data = useMemo(() => {
     if (isLoading || isError) return undefined
     return {
       resolver: getOffchainDnsResolverStatus({ chainId, dnsOffchainData }),
       address: checkDnsAddressMatch({
         address,
-        dnsAddress: addressRecord?.value as Address | undefined | null,
+        dnsAddress: dnsOffchainData?.extraData as Address | undefined | null,
```
talentlessguy, line 65: `dnsOffchainData.extraData` is what Cloudflare DNS returns from the DNS name. Not sure if that's correct, but `useAddressRecord` is always null

TateB: if the address value is resolving, it means the dns resolution is working correctly. if useAddressRecord is resolving to null, there's an issue with resolution

talentlessguy: I checked dns-query requests, they all return a correct value. but the hook keeps returning null. it doesn't return null only if address is already imported. and this issue happens both on mainnet and sepolia.

TateB:
ill merge and clean up the issues

PR #631, opened 2024-02-06
Viem test fix todos + reactivate ts & eslint 

storywithoutend:


sugh01 (approved):

src/hooks/primary/useGetPrimaryNameTransactionFlowItem/index.ts
```diff
@@ -101,19 +110,19 @@ export const useGetPrimaryNameTransactionFlowItem = (
           ? {
               resumeable: true,
               intro: {
-                title: [getIntroTranslation(introType, 'title'), { ns: 'transactionFlow' }],
+                title: [getIntroTranslation(introType, 'title'), { ns: 'transactionFlow' }] as any,
```
TateB, line 113: is as any required here

src/pages/register.tsx
```diff
@@ -51,7 +51,6 @@ export default function Page() {
     }
   }
 
-  // TODO: Temp fix for hydration issues.
```
TateB, line 54: can't we skeletonise this for hydration rather than return null

src/transaction-flow/input/RevokePermissions/views/SetExpiryView.tsx
```diff
@@ -96,6 +96,7 @@ export const SetExpiryView = ({
     year: 'numeric',
   })
 
+  console.log('formErrors', formState.errors.expiryCustom)
```
TateB, line 99: remove log

TateB (changes requested):

src/hooks/primary/useGetPrimaryNameTransactionFlowItem/index.ts
storywithoutend: Nope. Thanks for the catch. I forgot to delete it.

src/pages/register.tsx
storywithoutend: Removed the return null on loading as the passing the updated isLoading variable into <Registration/> seemed to stop the hydration errors.

PR #633, opened 2024-02-08
fix subgraph sync error bugs

storywithoutend:


talentlessguy (approved):

src/utils/useHasSubgraphSyncErrors.ts
```diff
@@ -13,79 +13,77 @@ type EventData = {
...
-    if (isError) {
-      if (isFailedToFetch) errorQueries += 1
-      else query.invalidate()
-    } else if (isSlow) {
-      slowQueries += 1
-    }
-  })
+      const isError = query.state.status === 'error'
 
-  return { slow: slowQueries, error: errorQueries }
```
talentlessguy, line 46: since it's change to booleans, why not return them as booleans? logic will stay the same in `useHasGlobalError`

PR #636, opened 2024-02-13
Update React Query to v5, RainbowKit, viem and wagmi to v2

talentlessguy:
TODO:

- [ ] Upgrade to React Query v5
- [ ] Upgrade to RainbowKit v2
- [ ] Upgrade to viem and wagmi v2
- [ ] Make sure there are no build errors
- [ ] Test out the app

PR #641, opened 2024-02-19
Wagmi/Viem/RainbowKit V2, React Query v5

talentlessguy:
Based on #636 but with Vitest for test runner so it's easier to transition

TODO:
- [x] make all tests pass
- [x]  make it build
- [x] fix multicall batch
- [x] use ensjs [with viem](https://github.com/ensdomains/ensjs-v3/pull/175) (currently needs yalc)


talentlessguy:
image

not sure if this is related to the PR but seems like name resolution now fails on anvil

talentlessguy:
For whatever reason subgraph fails locally

image

storywithoutend (approved):

PR #651, opened 2024-02-27
Fix grace period conditions

storywithoutend:
* Added support for grace period names
* Switched isWrapped to using !!wrapperData from ownerData.ownershipLevel === 'nameWrapper'
* Added error messages for canEditResolver + canCreateSubnames
* In Tokens.tsx, removed hook for getting fuse state and replaced with function using wrapperData
* When calculating expiry date from wrapperData.expiry, subtract the grace period time
* Replaced useBasicName(<parent-name>) with useParentBasicName so it will be easier to unit test

In general, unwrapped grace period names lose access to the registrar but keep access to the registry and resolver. For wrapped grace period names, lose access to the name wrapper contract.

A detailed list of outcomes grouped together can be found in the makeMockUseAbilitiesData.ts

|  Situation | canSendOwnership | canEditRecords | canEditResolver | canCreateSubnames |  
| --- | --- | --- | --- | --- |
| unwrapped-2ld | YES | YES | YES | YES | 
| wrapped-2ld | YES | YES | YES | YES |
|  grace-period-unwrapped-2ld | NO | YES | YES | YES | 
| grace-period-wrappe-2ld | NO | YES | NO | NO |


LeonmanRolls (approved):

src/hooks/abilities/useAbilities.test.ts
```diff
@@ -1,22 +1,45 @@
 import { mockFunction, renderHook } from '@app/test-utils'
+// import { writeFileSync} from 'fs'
```
TateB, line 2: remove

src/hooks/ensjs/subgraph/useSubgraphRegistrant.ts
```diff
@@ -0,0 +1,83 @@
...
+              id
+            }
+          }
+        }
+      }
+    `
+    const result = await subgraphClient.request<SubgraphResult>(query, { namehash: namehash(name) })
+    const registrant = result?.domain?.registration?.registrant?.id
+    if (!registrant) return null
+    return registrant
```
TateB, line 53: there's an ensjs func for this!

src/hooks/nameType/useNameType.ts
```diff
@@ -0,0 +1,44 @@
...
+export const useNameType = (name: string, options: Options = {}) => {
+  const enabled = options.enabled ?? true
+
+  const basicName = useBasicName({ name, enabled })
+  const nameWrapperAddress = useContractAddress({ contract: 'ensNameWrapper' })
+
+  const { isLoading, isCachedData } = basicName
+
+  const data: NameType | undefined = useMemo(() => {
+    if (isLoading) return undefined
```
TateB, line 20: is it necessary to do this if cached data state is enforced?

src/hooks/useNameDetails.test.ts
```diff
@@ -46,3 +47,389 @@ describe('useNameDetails', () => {
...
+      "resolverAddress": "0x0000000000000000000000000000000000000000"
+  },
+  "isLoading": false,
+  "isCachedData": false,
+  "registrationStatus": "available",
+  "type": "name",
+  "isShort": false,
+  "is2LD": true,
+  "isETH": true,
+  "labelDataArray": [
```
TateB, line 82: you can just spread the validation func result into this to remove some bloat here

src/hooks/useParentBasicName.ts
```diff
@@ -1,14 +1,9 @@
...
-  const validParentName = isValidParent ? parentName : ''
+import { useBasicName, UseBasicNameOptions } from '@app/hooks/useBasicName'
+import { parentName } from '@app/utils/name'
 
+export const useParentBasicName = ({ name, enabled, ...otherProps }: UseBasicNameOptions) => {
+  const parentName_ = parentName(name)
   // So far the use of this hook is for the wrapper data which does not rely on the graph so
   // it is safe to skip the graph
-  return useBasicName({ name: validParentName })
+  return useBasicName({ name: parentName_, ...otherProps })
```
TateB, line 8: since `subgraphEnabled` was readded to `useBasicName`, should that be specified as false here?

test/mock/makeMockUseContractAddress.ts
```diff
@@ -0,0 +1,57 @@
...
+  },
+  ensReverseRegistrar: {
+    address: deploymentAddresses.ReverseRegistrar,
+  },
+  ensBulkRenewal: {
+    address: deploymentAddresses.StaticBulkRenewal,
+  },
+  ensDnssecImpl: {
+    address: deploymentAddresses.LegacyDNSSECImpl,
+  },
```
TateB, line 44: not sure if this is actually used but this shouldn't use the legacy deployment address

test/mock/makeMockUseContractAddress.ts
```diff
@@ -0,0 +1,57 @@
...
+  legacyPublicResolver: {
+    address: deploymentAddresses.LegacyPublicResolver,
+  },
+  publicResolver: {
+    address: deploymentAddresses.PublicResolver,
+  },
```
TateB, line 50: dupe?

test/mock/makeMockUseValidate.ts
```diff
@@ -0,0 +1,179 @@
...
+  dns: { input: 'com' },
+  'valid-2ld': { input: 'name.eth' },
+  'valid-2ld:dns': { input: 'name.com' },
+  'invalid-2ld': { input: 'name❤️.eth' },
+  'valid-subname': { input: 'subname.name.eth' },
+} as const
+export type MockUseValidateType = keyof typeof mockUseValidateConfig
+export const mockUseValidateTypes = Object.keys(mockUseValidateConfig) as MockUseValidateType[]
+
+export const makeMockUseValidate = (type: MockUseValidateType): ValidationResult => {
```
TateB, line 17: does useValidate need to be mocked

TateB (changes requested):
minor nits

src/hooks/abilities/useAbilities.test.ts
storywithoutend: I have a script that groups the results together to make the results easier to read at the end that uses the writeFileSync, which is why I commented it out instead of removing it.

test/mock/makeMockUseValidate.ts
storywithoutend: In the absence of tests for the individual functions, imo I would say it useful as a test to make sure that the results remain consistent with what we expect in case we modify any of those functions in the future.

src/hooks/useParentBasicName.ts
storywithoutend: I removed the comment. This is no longer true anymore. replaced useBasicname(<parent-name>) with useParentBasicName so that it would be easier to mock the data in the future. I removed the comment.

test/mock/makeMockUseContractAddress.ts
storywithoutend: Originally copied the chain mapping from addTestContracts...  I refactored it so that the base data is coming from constants/chains.ts.

test/mock/makeMockUseContractAddress.ts
storywithoutend: same as above comment

src/hooks/useNameDetails.test.ts
storywithoutend: removed the mock from useNameDetails since it's not used at this time.

src/hooks/ensjs/subgraph/useSubgraphRegistrant.ts
storywithoutend: updated with ensjs function

src/hooks/ensjs/subgraph/useSubgraphRegistrant.ts
```diff
@@ -74,10 +56,15 @@ export const useSubgraphRegistrant = <TParams extends UseSubgraphRegistrantParam
...
+  const query = useQuery({
     ...preparedOptions,
     gcTime,
     enabled: enabled && !!params.name,
     staleTime,
   })
+
+  return {
+    ...query,
+    isCachedData: query.status === 'success' && query.isFetched && !query.isFetchedAfterMount,
```
TateB, line 68: use `getIsCachedData`

TateB (approved):
sync from `viem-test`, then lgtm

PR #668, opened 2024-03-06
fix: NFT image not loading because of CSP errors

talentlessguy:


talentlessguy:
It loads locally but in staging env it doesn't show up:

image

TateB (approved):

PR #673, opened 2024-03-08
fix: hide expiry buttons if out of grace period

talentlessguy:
image


src/utils/registrationStatus.ts
```diff
@@ -62,7 +62,7 @@ export const getRegistrationStatus = ({
       if (expiry.getTime() + gracePeriod * 1000 > timestamp) {
         return 'gracePeriod'
       }
-      const { premium } = priceData!
+      const { premium } = priceData || { premium: 0n }
```
talentlessguy, line 65: this throws a client error if you open an expired name (not caused by me)

LeonmanRolls (approved):

sugh01 (approved):

PR #680, opened 2024-03-11
Viem test+main

storywithoutend:
List of files that had merges that included both main and viem-test updates:
- useSimpleSearch
  - removed one of the queires 
- AvatarNFT
  -  Kept the NftItem component and LoadFailureContainer
- ExpirySection
  -  The merge for some reason had questions of if the earnifi dropdown should be on the remind me or extend button. Might have been my own mistake since I had to add the updated earnifi => bankless translations for both the dropdown and the modal title.
- useMoonpayRegistration
  - kept encodeURIComponent(normalisedName) in the requestUrl
- DNSClaim
  - Not sure if I selected wrong option but file was listed as deleted in uncommitted changes after the initial merge.


src/utils/getDefaultWallets.ts
```diff
@@ -26,6 +27,8 @@ export const getDefaultWallets = ({
         injectedWallet,
         safeWallet,
         braveWallet,
+        // TODO: Double check phantom wallet
+        phantomWallet,
```
TateB, line 31: didn't we have a custom phantom wallet

TateB (changes requested):

PR #681, opened 2024-03-12
fix: loading optimisations

TateB:
changes:
- `useRecords`
  - added placeholder data function which can get data from prior queries to show initial data on load
  - added prefetch hook
- added `useSubgraphRecords` prefetch
- added `useProfile` prefetch
- changed addr to always load, which allows CCIP-read names to load an initial page significantly quicker
- added profile prefetch in searchresult


LeonmanRolls (changes requested):
Separate out functions in useRecords and add unit tests

sugh01 (approved):

LeonmanRolls (approved):

PR #709, opened 2024-04-16
update fallback chains for wildcard resolution

aslobodnik:
**Notes**
1. This increases total number requests by 1
2. Added polygon support because of this feature [request](https://ens.canny.io/feature-request/p/add-polygon-ccip-support) on canny
3. Introduces a bug that cb.id names no longer display eth address, see example below
How cb.id names would look after merge 
image
How cb.id names look currently
image


TateB:
if you remove ltc as well then we stay at the same amount of requests

aslobodnik (commented):
removed ltc per tate's feedback

LeonmanRolls:
Thanks @aslobodnik , closing as this was updated here: #706

PR #736, opened 2024-05-30
Update to react query v5

storywithoutend:
**Fixes**
* Flashing banner on profile page load.
* Update useQuery and useInfiniteQuery for react-query v5
* Warning from adding keys to components through spread operator. Disallowed in React 18.

**Notes**
I used a wrapper hook around useQuery to get **isLoading** to behave the way we expect it to. In v5 isLoading is only true if it is the first load and useQuery is fetching. If useQuery uses cached data, isLoading will never be true. In v5, isPending was supposed to replace isLoading, but isPending will remain true if useQuery is disabled, where before isLoading would be false. The solution was to alter the behavior of isLoading by making it true when the hook is enabled & isPending is true.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "@tanstack/query-sync-storage-persister": "5.22.2",
     "@tanstack/react-query": "5.22.2",
     "@tanstack/react-query-persist-client": "5.22.2",
-    "@wagmi/core": "^2.6.16",
+    "@wagmi/core": "2.6.16",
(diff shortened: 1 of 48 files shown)


LeonmanRolls (approved):

sugh01 (approved):

sugh01:
Found a dotbox issue, please dont merge yet

sugh01 (approved):

PR #737, opened 2024-05-31
Add support to Cardano ADA on user Profile for Offchain resolver

francisluz:
Hi Team,

I'm from Begin Wallet currently supporting Cardano and we're launching https://beginid.io/ based on ENS offchain resolver. and we'd like to list the ADA address when users come to the ENS app.


LeonmanRolls:
Hi, unfortunately we can't add ADA at this time as we are already at the limit of the number of requests we would like to send out when loading the profile page. We are addressing this problem behind the scenes though.

PR #744, opened 2024-06-16
fix: github username regex

o-az:
I tried updating my ENS profile on https://app.ens.domains to add an email and received a github username error even though I already had the username added to my profile (https://ens.app/esm.eth):

CleanShot 2024-06-16 at 08 18 24

therefore I'm assuming this is a regression bug. So I updated the GitHub username regex and added another test case.

New regex:

```sh
/^[a-z\d](?:[a-z\d]|-(?=[a-z\d])){0,38}(-?)?$/i
```

regex testing playground: https://regexr.com/81nk2


talentlessguy:
seems like the regex can match empty strings, but I might be wrong

o-az:
> seems like the regex can match empty strings, but I might be wrong

I added an empty string check here to show it doesn't match

regexr.com/81nk2

I'll add an empty string test case too

o-az:
added empty string github username test case

PR #745, opened 2024-06-19
Add guard condition before getCurrentTransaction()

storywithoutend:
** TLDR **
1. Start a transaction flow
2. Click "Open Wallet"
3. Close the transaction modal
4. Approve the transaction in the wallet
5. App crashes

** Diagnosis **
When the user closes the modal the transaction store runs the events **stopFlow** and **cleanupTransaction**. This nulls out the selectedKey and deletes the transaction from the items dictionary. When the wallet approves the transaction, it dispatches the action **setTransactionHash**, but the reducer cannot find the current transaction since is has been cleared from memory, causing an error to occur getCurrentTransaction attempts to access a property on undefined.

** What to also consider before approving **
I looked through to see if there were any side effects from not setting the hash in the transaction flow. I couldn't find any as it appears that when the transaction is successful, it is added to recent transactions and then other operations involving the hash of the transaction are independent from the transaction store.


TateB (approved):

sugh01 (approved):

sugh01 (changes requested):
Need to double check the "Commitment too Old" situation, where user gets stuck on the transaction modal

sugh01 (approved):

PR #746, opened 2024-06-25
Dep cleanup

talentlessguy:
This PR cleans up a lot of uneeded / unused deps, to speed up CI times and local project spin up speeds.

* remove ganache
* use globby instead of glob (25kb vs 475kb)
* remove direct fetch deps (node.js 18 supports fetch ootb)
* remove other misc deps that aren't used anywhere


talentlessguy:
I'll instead cherry pick the commits for #658

PR #769, opened 2024-07-18
fixing date picker choosen date for other timezones

yeewiz:


storywithoutend (changes requested):
Can we move this fix from the *roundDurationWithDay* to *Calendar.tsx*?  I think it makes more sense there with a comment. 

`onChange={(e) => {
// Update e to have valueAsDate with a localized value
props.onChange?.(e)
}}`

yeewiz:
> Can we move this fix from the _roundDurationWithDay_ to _Calendar.tsx_? I think it makes more sense there with a comment.
> 
> `onChange={(e) => { // Update e to have valueAsDate with a localized value props.onChange?.(e) }}`

@storywithoutend should be moved into that component

sugh01 (approved):

PR #775, opened 2024-07-24
FET-1571 - Addresses on profile page should have dropdowns with additional actions

yeewiz:
- Added the ability for the address profile button to be a dropdown with options instead of just a copy button
- Added additional unit tests for the address profile button
- Added a new hook useCoinChain to determine the chain by the coin type


scripts/coin-mapper.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// convert these into objects
+const supportedBlockExplorers = supportedCoins.map((address: string) => {
+  const coinType = coinNameToTypeMap[address as CoinName]
+  if (address === 'btc') return btcBlockExplorer
+  if (address === 'sol') return solBlockExplorer
+
+  // Ethereum mainnet is 60, but we need to convert it to 1
+  const coinId = coinType === 60 ? 1 : coinTypeToEvmChainId(coinType)
```
LeonmanRolls, line 65: does `coinTypeToEvmChainId` fail to handle ethereum mainnet?

storywithoutend: Yes. All it does is remove the leading bitwise 1 and then convert the result to a number.  I having a feeling it will fail with **etc** as well.

src/hooks/chain/useCoinChain.ts
```diff
@@ -0,0 +1,107 @@
...
+  const lowerCoinName = coinName?.toLowerCase() || ''
+  const coinType = coinNameToTypeMap[lowerCoinName as CoinName]
+
+  if (supportedAddresses.includes(coinName?.toLowerCase() as SupportedAddress)) {
+    const supportedBlockExplorers = await import('../../constants/blockExplorers/supported.json')
+    const coinBlockExplorer = supportedBlockExplorers?.default.find(
+      (blockExplorer) =>
+        (isEvmCoinType(coinType) &&
+          blockExplorer?.id === coinTypeToEvmChainId(coinType as EvmCoinType)) ||
+        blockExplorer?.nativeCurrency?.symbol.toLowerCase() === lowerCoinName, // this is needed for certain coins such as eth, btc, sol, etc.
```
storywithoutend, line 61: *nativeCurrency.symbol* does not always match coinName. coinName is a symbol that we created to make it more user friendly to when working with multiple chains. So instead of seeing numbers like 60, 1093234, etc, we would just see eth, base, etc.

See comment in supported.json on how we should handle this.

src/constants/blockExplorers/supported.json
```diff
@@ -0,0 +1 @@
...urrency":{"name":"Bitcoin","symbol":"BTC","decimals":8},"blockExplorers":{"default":{"name":"Joseon Explorer","url":"https://www.blockexplorer.com/"}}},{"name":"Solana","nativeCurrency":{"name":"Solana","symbol":"SOL","decimals":9},"blockExplorers":{"default":{"name":"Solana","url":"https://explorer.solana.com/"}}},{"id":10,"name":"OP Mainnet","nativeCurrency":{"name":"Ether","symbol":"ETH","decimals":18},"blockExplorers":{"default":{"name":"Optimism Explorer","url":"https://optimistic.etherscan.io","apiUrl":"https://api-optimistic.etherscan.io/api"}}},{"id":42161,"name":"Arbitrum One","nativeCurrency":{"name":"Ether","symbol":"ETH","decimals":18},"blockExplorers":{"default":{"name":"Arbiscan","url":"https://arbiscan.io","apiUrl":"https://api.arbiscan.io/api"}}},{"id":8453,"name":"Base","nativeCurrency":{"name":"Ether","symbol":"ETH","decimals":18},"blockExplorers":{"default":{"name":"Basescan","url":"https://basescan.org","apiUrl":"https://api.basescan.org/api"}}},{"id":137,"name":"Polygon","nativeCurrency":{"name":"MATIC","symbol":"MATIC","decimals":18},"blockExplorers":{"default":{"name":"PolygonScan","url":"https://polygonscan.com","apiUrl":"https://api.polygonscan.com/api"}}}]
```
storywithoutend, line 1: We should use a dictionary here.

{ [key: coinName]: blockExplorerData }

Using coinName for the key will make it easier to find the info and remove the need to use the *find* method.

storywithoutend (changes requested):

src/constants/blockExplorers/supported.json
yeewiz: made this update to supported.json, but also the evm.json file by modifying the coinmapper

src/hooks/chain/useCoinChain.ts
yeewiz: updated logic to use the coinName to get the blockexplorer for both supported and evm coins

PR #781, opened 2024-07-26
FET-1335: Make ETH record permanently visible when editing profile

staslysak:


src/components/pages/profile/[name]/registration/steps/Profile/profileRecordUtils.ts
```diff
@@ -240,6 +246,8 @@ export const getProfileRecordsDiff = (
     .filter((r) => !!r) as ProfileRecord[]
   const deletedRecords = previousRecords
     .filter((previousRecord) => {
+      if (isEthAddressRecord(previousRecord)) return true
+
```
Comment (storywithoutend) [inline src/components/pages/profile/[name]/registration/steps/Profile/profileRecordUtils.ts:250]: This causing a bug. It will delete any eth value if it exists in the previous record

LeonmanRolls (approved):

src/transaction-flow/input/ProfileEditor/ProfileEditor-flow.tsx
```diff
@@ -275,7 +277,13 @@ const ProfileEditor = ({ data = {}, transactions = [], dispatch, onDismiss }: Pr
                     validator={validatorForRecord(field)}
                     validated={isDirtyForRecordAtIndex(index)}
                     error={errorForRecordAtIndex(index, 'key')}
-                    onDelete={() => handleDeleteRecord(field, index)}
+                    onDelete={() => {
+                      if (isEthAddressRecord(field)) {
+                        updateRecordAtIndex(index, { ...field, value: '' })
+                      } else {
```
storywithoutend, line 283: This is unnecessary since the field group is "custom" which means that it will always be a text record.

src/transaction-flow/input/ProfileEditor/ProfileEditor-flow.tsx
```diff
@@ -286,7 +294,13 @@ const ProfileEditor = ({ data = {}, transactions = [], dispatch, onDismiss }: Pr
                     placeholder={placeholderForRecord(field)}
                     error={errorForRecordAtIndex(index)}
                     validated={isDirtyForRecordAtIndex(index)}
-                    onDelete={() => handleDeleteRecord(field, index)}
+                    onDelete={() => {
+                      if (isEthAddressRecord(field)) {
+                        updateRecordAtIndex(index, { ...field, value: '' })
+                      } else {
```
storywithoutend, line 300: This is unnecessary since the key is "descriptions" which means it will be a <textarea/> tag. Which will never be an address.

storywithoutend (changes requested):

PR #788, opened 2024-08-01
Move extend-names logic to ExtendNames-flow

LeonmanRolls:


pnpm-lock.yaml
```diff
@@ -451,6 +451,39 @@ importers:
         specifier: ^1.0.0-pre.53
         version: 1.0.0-pre.53
 
+  .yalc/@ensdomains/ensjs:
```
talentlessguy, line 454: i think you forgot to unlink local version of ensjs. pls remove

talentlessguy (changes requested):

PR #806, opened 2024-08-09
fix: detect desynced expiry + show banner

TateB:
relies on thorin pr: https://github.com/ensdomains/thorin/pull/156

main issue:
- a 2ld .eth name can have differing expiries between registrar and namewrapper (namewrapper can be lower)
- if namewrapper expiry has passed, ownership shows as 0x0 because the name is still at the namewrapper level
- in this situation, various parts of the app break because they assume it's not possible, and the name becomes effectively locked (from the app's context)
- the user has no recourse for fixing it

fix:
- get ownership value regardless of expiry
  - allows showing the owner so the user doesn't panic, and so the rest of the UI doesn't behave too weirdly
- add new nametype for this specific desync situation
  - with no ability for any actions (i.e. send/edit)
- show banner for desynced name with fix button
  - fix button is the same as just renewing the name for 0 duration (with 0 value)

changes:
- `useOwner` now has a `forceUnexpired` parameter which allows fetching ownership regardless of expiry (for both wrapper+registrar)
  - relies on a bytecode override for both the registrar and the namewrapper with a simple fetcher that can be found here: https://gist.github.com/TateB/d84e4789e1419a637fda6e23250fa1df
  - at the app level rather than ensjs because it's context specific and requires node support
- the `useOwner` call in `useBasicName` now fetches ownership regardless of expiry
  - not sure if this has unintended consequences, but we were doing something kinda similar for grace period anyway
- new `Misconfigured name` banner with an button for repair if a name is detected as having a desynced expiry
- add `syncNameExpiry` transaction
- updated styled-components/react-transition-state (from thorin req)


laurgk:
Fixed by Dave on a diff PR

PR #812, opened 2024-08-13
Configure Knip integration and remove unused files from src.

nhohb:


LeonmanRolls (approved):

sugh01:
- Please add a section for `knip` to readme, like steps to `configure` and `run` knip 
- Let us add a workflow for `knip` to run on CI

PR #814, opened 2024-08-13
use @ensdomains/headless-web3-provider

talentlessguy:
Don't use ethers in tests, use our own fork of headless-web3-provider


.github/workflows/pages-deployment.yaml
```diff
@@ -45,7 +45,7 @@ jobs:
       - name: Install Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: 20
```
talentlessguy, line 48: to be able to use `fs.glob` from Node v20

sugh01 (approved):

PR #818, opened 2024-08-15
Fet 1610 merge from main

storywithoutend:


playwright/fixtures/accounts.ts
```diff
@@ -17,47 +16,47 @@ const shortenAddress = (address = '', maxLength = 10, leftSlice = 5, rightSlice
   return `${address.slice(0, leftSlice)}...${address.slice(-rightSlice)}`
 }
 
+export type Dependencies = {
```
talentlessguy, line 19: no need to depend on a provider, use `testClient` from test contracts file directly

talentlessguy (changes requested):

PR #835, opened 2024-08-28
FET-1619 If tab specified in query string doesn't exist, app crashes. Should redirect to profile tab

vuongnmdevblock:


nhohb (approved):

src/components/pages/profile/[name]/Profile.tsx
```diff
@@ -177,13 +177,16 @@ const ProfileContent = ({ isSelf, isLoading: parentIsLoading, name }: Props) =>
   // profile.decryptedName fetches labels from NW/subgraph
   // normalisedName fetches labels from localStorage
   useEffect(() => {
+    if (!tab || !tabs.includes(tab)) {
+      router.replace(`/${normalisedName}`)
+    }
     shouldRedirect(router, 'Profile.tsx', '/profile', {
```
Comment (storywithoutend) [inline src/components/pages/profile/[name]/Profile.tsx:183]: Can you move the redirect logic into `shouldRedirect.tx`?  That way all redirect logic is in the same area.

storywithoutend (changes requested):

src/components/pages/profile/[name]/Profile.tsx
```diff
@@ -182,8 +182,10 @@ const ProfileContent = ({ isSelf, isLoading: parentIsLoading, name }: Props) =>
       name,
       decodedName: profile?.decodedName,
       normalisedName,
+      tabs,
+      tab,
```
Comment (storywithoutend) [inline src/components/pages/profile/[name]/Profile.tsx:186]: Pass in "visibleTabs" instead of just tabs. "visibleTabs" will filter out the permissions tab for legacy names.  This may cause a dependency array error. Let me know if it does and I can help you fix it.

src/utils/shouldRedirect.ts
```diff
@@ -97,7 +99,7 @@ export const shouldRedirect = (
       ['Profile.tsx', { isSelf: P.boolean, name: P.string }],
       (_params) => _params && router.isReady,
       (_params) => {
-        const { name, isSelf, decodedName, normalisedName } = _params[1]
+        const { name, isSelf, decodedName, normalisedName, tabs, tab } = _params[1]
 
         if (
           name !== decodedName &&
```
storywithoutend, line 105: Let's add the rerouting code to all of the conditions here. e.g.

const hasValidTab = tabs.includes(tab)
const tabQuery = tab !== 'profile' && hasValidTab ? `?tab=${tab}` : ''

... .... 
router.replace(`${destination}/${decodedName}${tabQuery})


... ...
if (!hasValidTab) router.replace(...)

storywithoutend (changes requested):
Please don't forget to add tests to shouldRedirect.test.ts

src/utils/shouldRedirect.ts
vuongnmdevblock: I encountered an infinite loop when attempting to reroute the valid tab query using the shouldRedirect method. To resolve this, I temporarily commented out the `return router.replace(destination/{name}${tabQuery})` code. The issue is resolved when removing the `router` dependency. I would appreciate your guidance on addressing this infinite loop.

storywithoutend (approved):

PR #840, opened 2024-08-30
Add registration tracking events for .eth and TLD (.box) funnel

nhohb:
- Add tracking events to registration flow
- Tracking events should only be called once
- Transaction modal tracking events should only be called for the specific transaction
- Refactor trackEvents so that it can accept a payload and is typed


next.config.mjs
```diff
@@ -23,7 +23,7 @@ const babelIncludeRegexes = [
  * @type {import('next').NextConfig}
  * */
 const nextConfig = {
-  reactStrictMode: true,
+  reactStrictMode: process.env.NODE_ENV !== 'development',
   compiler: {
```
storywithoutend, line 27: Did you turn off strict mode because it renders components twice?

src/components/@molecules/TransactionDialogManager/stage/TransactionStageModal.tsx
```diff
@@ -476,7 +477,12 @@ export const TransactionStageModal = ({
           !!requestError ||
           isTransactionRequestCachedData
         }
-        onClick={() => sendTransaction(request!)}
+        onClick={() => {
+          sendTransaction(request!)
+          trackRegistrationEvent(
+            actionName === 'commitName' ? 'Commit Wallet Opened' : 'Finish Wallet Opened',
```
storywithoutend, line 483: You will need to add a condition before this or it will send a track event for every transaction.

next.config.mjs
nhohb: @storywithoutend yes it's to prevent the useEffect running twice

src/pages/index.tsx
```diff
@@ -92,6 +94,12 @@ export default function Page() {
   const { t } = useTranslation('common')
 
   const { dialogProps } = useVerificationOAuthHandler()
+  const { trackRegistrationEvent } = useRegistrationEventTracker()
+
+  useEffect(() => {
+    trackRegistrationEvent('Home page')
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
```
storywithoutend, line 103: I think this is why you set strictMode to false. In dev mode it will run twice but in production it will only run once. This is okay.

storywithoutend (changes requested):

src/components/@molecules/SearchInput/SearchInput.tsx
```diff
@@ -338,6 +341,8 @@ const createSearchHandler =
       { lastAccessed: Date.now(), nameType, text, isValid: selectedItem.isValid },
     ])
 
+    trackRegistrationEvent('Search selected', { keyword: text })
```
makoto, line 344: Since existing event names are lower cased (`register/commit/renew`) shall we do all snake case? `search_selected` not `Search selected` ?

nhohb: @makoto  I've switched to snake case. Please review the changes, thank you.

storywithoutend (changes requested):
2 tests to add: 

- e2e/specs/stateful/box.spec.ts: Add a new file that follows the dot box registration and fires the tracking events.
- e2e/specs/stateless/registerName [line 160] add a check that **search_selected_eth** is not fired from this search input.

Also. I know we haven't done this a lot but the registerName tests are starting to get really big so can you wrap your **expect** code with **test.step('should fire tracking event: <event>', () => {...}**?

src/components/@molecules/SearchInput/SearchInput.tsx
```diff
@@ -338,6 +341,13 @@ const createSearchHandler =
...
     ])
 
+    if (nameType === 'eth' || nameType === 'box') {
+      trackEvent({
+        eventName: `search_selected_${nameType}`,
+        customProperties: { name: text },
+      })
+    }
+
     const path = getRouteForSearchItem({ address, chainId, queryClient, selectedItem })
```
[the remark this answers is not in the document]
nhohb: @storywithoutend
I've implemented your suggestion, but I'm encountering an issue. When a user enters a keyword and immediately presses Enter, the path always resolves to `/profile/{name}` instead of `/register/{name}`.

I believe the `ownerData` variable is `undefined` because `getCachedQueryData` isn't returning data in time.

To address this, I've added a `waitForTimeout` in `e2e/specs/stateless/registerName.spec.ts` and included a comment explaining the reason.

Please review these changes and let me know if you have any further feedback.

src/components/pages/profile/[name]/registration/steps/Transactions.tsx
```diff
@@ -140,6 +142,8 @@ const Transactions = ({ registrationData, name, callback, onStart }: Props) => {
       autoClose: true,
       resumeLink: `/register/${name}`,
     })
+
+    trackEvent({ eventName: 'register_started' })
   }
```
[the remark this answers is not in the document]
Comment (nhohb) [reply src/components/pages/profile/[name]/registration/steps/Transactions.tsx:147]: I've analyzed the flow and determined the following:

- If a user completes a commit transaction, the `registerTx` variable will be `undefined`.
The `makeRegisterNameFlow` function will be called.

- If the user refreshes the page, the `registerTx` variable will have a value, and the `showRegisterTransaction` function will be called.

src/utils/analytics.ts
```diff
@@ -31,15 +31,15 @@ export function getUtm() {
 export const setupAnalytics = () => {
   setUtm()
 }
-
-export const trackEvent = async (type: string, chain: string) => {
+export const trackEvent = async (type: string, chain: string, customProperties?: any) => {
   const referrer = getUtm()
+
+  const props = referrer ? { ...customProperties, referrer } : customProperties
```
storywithoutend, line 37: I don't think we need *referrer ? {} : {}*, we can just pass { ...customProperties, referrer}

storywithoutend (changes requested):

storywithoutend (approved):

LeonmanRolls (approved):

sugh01 (approved):

PR #885, opened 2024-09-26
FET-1661: Address/My names page click through not working

staslysak:


sugh01:
Fixed with another pr

PR #929, opened 2025-01-16
use new ens-test-env

talentlessguy:
uses gzip instead of lz4 - saves 8-11 seconds in CI on each stateful and stateless test during the `pnpm rebuild -r` step


package.json
```diff
@@ -179,7 +179,7 @@
     "vite-plugin-magical-svg": "^1.3.0",
     "vitest": "^2.0.5",
     "vitest-canvas-mock": "^0.3.3",
-    "wait-on": "^6.0.1",
+    "wait-on": "^8.0.2",
```
talentlessguy, line 182: this is to dedupe `wait-on` dependency

sugh01 (approved):

PR #955, opened 2025-02-14
ci: optimise ci build time

nexxeln:


.github/workflows/test.yaml
talentlessguy, line 1: the reason why it failed with arm64 is because CI would pull the amd64 and not arm build of chrome. The problem is that ARM builds are not hosted officially by Chrome or MS, but there are community-made builds:

https://github.com/chromium-for-lambda/chromium-binaries/releases/tag/x86_64-amazon-linux-2023-chromium-127.0.6533

nexxeln: is it worth using a community-made build?

PR #980, opened 2025-04-02
fix: safe (or other sca) gas estimates

TateB:
most (if not all) SCA implementations use more than the gas allocated for a simple `.transfer()`. this means that when interacting with a contract that utilises `.transfer()`, that function will fail with an `out of gas` error by default. to avoid this, we can generate an access list of storage slots accessed during the transaction, which then pays for the cold gas of any storage slot at the start of the transaction. with cold gas paid for, during the execution of `.transfer()`, only the warm gas is paid, and (generally) this means it stays below the gas allowed for that operation.

access lists were already being generated for safe transactions in the transaction dialog, but this PR expands that to any transaction (e.g. for non-safe SCAs), and adds support for them to be generated when doing register/renew gas estimates.

changes:
- estimating gas in the transaction dialog now generates an access list for all transactions, even if not connected via safe. this allows for compatibility with any SCA.
- when estimating gas with state overrides, a `.transfer` access list is now generated from running [this yul code](https://gist.github.com/TateB/777287c9a63d5f02fcd905232ce5748a). this just executes a transfer to the target SCA address (the address in the gist is just a placeholder), meaning the required access list for a `.transfer()` can be derived.

fixes fet-1800, fet-1799, fet-1771, fet-1772


LeonmanRolls (approved):

laurgk (approved):
Tested - good to go.

PR #985, opened 2025-04-11
chore: para rainbowkit version bump and set recovery step

jlm0:
Bumps `@getpara/rainbowkit` and @getpara/rainbowkit-wallet` to latest `1.2.0` release. Adds the `recoverySecretStepEnabled` flag to the `ParaModalProps` for new user creation.

TODO: 
* After release, check to make sure that para is on production mode. API key has been set to only be active on main branch

fixes fet-1822

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,1 @@
-* @sugh01
+* @sugh01 @laurgk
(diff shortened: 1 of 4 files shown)


pnpm-lock.yaml
```diff
@@ -9052,6 +9045,11 @@ snapshots:
       ethereum-cryptography: 2.2.1
       micro-ftch: 0.3.1
 
+  '@ethereumjs/util@9.1.0':
```
talentlessguy, line 9048: why is there a new dependency?

laurgk (approved):
Tested and good to go.

PR #995, opened 2025-04-25
fixed a stateful test

sugh01:


laurgk (approved):
Worked through and confirmed together :)

PR #998, opened 2025-04-29
fix: metamask explicit transaction type

TateB:
adds explicit transaction type for MetaMask transactions to avoid this issue: https://github.com/MetaMask/core/issues/5720

only added on MetaMask txs in case explicit type is an issue in any wallets


laurgk (approved):
Tested - looks good.

PR #1003, opened 2025-05-05
Preview thorin updates

storywithoutend:
* My names Selector label is now clickable
* Desktop toast is solid background in both light and dark mode
* Mobile toast is blurred background and can now be dragged up to hide or dragged down to expand.


laurgk (approved):
Tested - good to go.

PR #1024, opened 2025-06-23
Add safe wallet test

sugh01:


storywithoutend (changes requested):
I would just add some more constant logs, especially for error states so that we can get a better idea of where anything is failing when it's failing.

PR #1030, opened 2025-07-17
Fix cb sca

talentlessguy:
try to fix it by resetting local storage keys


laurgk:
cb issue fixed - closing for now

PR #1075, opened 2025-10-29
update support article links

davidullo:
Previous links led users to non-existent pages. Those have been corrected to point to the new support pages.


laurgk:
This has been updated - thank you

PR #1100, opened 2025-12-15
Fix flaky verifications test

storywithoutend:
The verifications test was flaky because navigating with OAuth params before connecting could show a "not logged in" error dialog that blocks the connect button. Now we connect first, then navigate with OAuth params.


laurgk:
Old PR with removed items that conflict with main (mostly Dentity).
Any important test changes will be raised in a separate PR.

PR #1110, opened 2026-02-07
fix: broken url to support page

nya-elimu:
Fix broken URL shown when updating records at `https://app.ens.domains/<name>.eth?tab=records`

https://app.ens.domains/elimu.eth?tab=records

> Screenshot 2026-02-07 154142

Points to non-existing page: https://support.ens.domains/en/articles/7982906-long-running-transactions

> Screenshot 2026-02-07 154241


laurgk:
This has been updated - thank you!

PR #1125, opened 2026-04-28
fix: use calendar-aware year calculation in renew modal to prevent infinite loop

### LINKED ISSUE #1076: Year input causes infinite increment in renew modal for large values
## Summary
In the renew modal’s years view, entering a large year (e.g., 1600) makes the displayed “xxx years extension” keep increasing on each render.

## Steps to Reproduce
1. Open the Extend modal and switch to the years view
2. Enter 1600 (or larger) into the numeric input
3. Observe the label and description increasing continuously

## Root Cause
The accumulated error from leap days for large year inputs causes the displayed year count to enter an increasing loop.

https://github.com/user-attachments/assets/1687d3fb-d0d0-4489-8af1-66304d105d79

tenderdeve:
## Summary
- Fix infinite increment bug in the renew modal when entering large year values (e.g. 1600)
- **Root cause:** Feedback loop between a leap-day-ignorant seconds-to-years conversion (`Math.floor(seconds / ONE_YEAR)` where `ONE_YEAR = 365 * 24 * 3600`) and a leap-day-aware years-to-seconds conversion (`Date.setFullYear()`). For large values, the leap day accumulation caused the year count to keep increasing on each render.
- **Fix:** Replace the naive `secondsToYears` floor division with `calculateDatesDiff(startDate, endDate).diff.years`, which uses calendar-aware date arithmetic consistent with the years-to-seconds path, creating a stable roundtrip.

**File:** `src/components/@molecules/DateSelection/DateSelection.tsx`

Fixes #1076

## Test plan
- [ ] Enter year value of 1600 in the renew modal — should not keep incrementing
- [ ] Enter normal values (1-10 years) — should work as before
- [ ] Switch between year and date views — should remain consistent

### DIFF
--- a/src/components/@molecules/DateSelection/DateSelection.tsx
+++ b/src/components/@molecules/DateSelection/DateSelection.tsx
@@ -6,9 +6,9 @@ import { Typography } from '@ensdomains/thorin'
 
 import { Calendar } from '@app/components/@atoms/Calendar/Calendar'
 import { PlusMinusControl } from '@app/components/@atoms/PlusMinusControl/PlusMinusControl'
-import { roundDurationWithDay, secondsFromDateDiff } from '@app/utils/date'
+import { calculateDatesDiff, roundDurationWithDay, secondsFromDateDiff } from '@app/utils/date'
 import { isInsideSafe } from '@app/utils/safe'
-import { formatDurationOfDates, secondsToYears } from '@app/utils/utils'
+import { formatDurationOfDates } from '@app/utils/utils'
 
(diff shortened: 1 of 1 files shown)


LeonmanRolls:
Closing in favour of #1130, which is a re-push of this exact commit from a maintainer branch so CI can run for first-time contributors. Original commit authorship preserved via cherry-pick, so you remain the commit author on the re-submission. Thanks again for the diagnosis and fix, @tenderdeve.

PR #1126, opened 2026-05-11
Add pcc subname expiry extension support

sugh01:


Claus36868:
**Heads-up (surfaced by the stacked WEB-110 PR #1152, routed here per its review).** #1152's CI is the first to actually execute this branch's stateless e2e — on this branch the `stateless (N)` shards all die at env setup (`ensindexer | Cannot build subgraph _Meta_ ZodError` → `Cannot read properties of null (reading 'block')`), so these tests never ran in this PR's own CI.

When the e2e env does come up, `e2e/specs/stateless/ownership.3LD.spec.ts:1624-1625` ("Wrapped & Emancipated 3LD … Parent Owner and Manager › Expiry Section, Extend & Set Reminder") fails:

```
await expect(ownershipPage.extendButton).toHaveCount(0)   // actual: 1
await expect(ownershipPage.setReminder).toHaveCount(0)     // actual: 1
```

Scenario: emancipated subname `test.wrapped.eth` (`PARENT_CANNOT_CONTROL` burned, `CAN_EXTEND_EXPIRY` **not** burned), owner `user`; parent `wrapped.eth` owner `user2`. The test logs in as the default `user` (the *subname* owner — `playwright/fixtures/login.ts` `connect(user = 'user')`).

The Extend / Set-reminder actions now render because `useExpiryActions` returns set-reminder unconditionally and `abilities.data?.canExtend` is true here. But for the subname owner without `CAN_EXTEND_EXPIRY` (and who isn't the parent owner), `NameWrapper.extendExpiry` would revert. So either `canExtend` / `isSelfExtendable` should also gate on `CAN_EXTEND_EXPIRY` (or parent-ownership), or this e2e expectation is intentionally obsolete and should be updated — your call which. Flagging because this branch's CI masked it via the env failure. (No exploit path: the contract enforces the real auth; worst case is a reverting tx / wasted gas.)

WEB-110 (#1152) does not touch this code (abilities / ownership tab / registration) — it's the cap/Max work in the extend modal only — so it can't and shouldn't fix this here.

Claus36868:
**Correction / refinement to my note above after a closer read — please disregard the "canExtend should gate on CAN_EXTEND_EXPIRY" part, it already does:**

- `canExtendPccSubname` (`src/hooks/abilities/useAbilities.ts:107-113`) **does** gate correctly: `(CAN_EXTEND_EXPIRY && isSelfExtendable) || isParentOwner`. For the failing scenario (subname owner, `CAN_EXTEND_EXPIRY` not burned, not the parent owner) it correctly returns `false`. So the gating logic isn't the bug — my earlier framing was off.
- The **deterministic** half of the `ownership.3LD` failure is `setReminder` (line 1625): `useExpiryActions` returns the set-reminder action **unconditionally** (whenever an expiry exists), so `toHaveCount(0)` fails for any name reaching that section — a behavior change on this branch. Decide whether set-reminder should be gated for this case or the e2e expectation updated.
- The `extendButton` also showing (count 1) *despite* the correct gating, **plus** the flaky `registerName` price/gas failure on the rerun, both coincide with this branch's `ensindexer | Cannot build subgraph _Meta_ ZodError: "Too small: expected number to be >=1"` → `TypeError: Cannot read properties of null (reading 'block')` at e2e env setup. **`main`'s latest Test run (2026-06-29) had all 32 stateless shards green**, so this subgraph-indexer build break looks **specific to this branch** (its own changes, or a `main` commit it rebased onto). That's the likely dominant root cause — it would explain data-dependent tests (ownership `canExtend`, registration price) misbehaving non-deterministically.

Suggested order: fix the branch's `ensindexer`/subgraph-build break first, then re-assess the `set-reminder`/`ownership.3LD` expectation. No security/exploit path (on-chain `extendExpiry` enforces the real auth; worst case is a reverting tx). None of this is WEB-110/#1152, which doesn't touch these files.

PR #1138, opened 2026-06-11
Issue 10 simplex profile category

brenzi:
closes #10, #11

List SimpleX not under social but its own priority category
Restrict number of urls to 1 and replace separator with semicolon


brenzi:
wrong target

PR #1139, opened 2026-06-11
configure redeployed namewrapper

brenzi:
This fixes https://github.com/simplex-network/ens-contracts/pull/18


brenzi:
wrong target

PR #1147, opened 2026-06-25
log cleanup

brenzi:
downgrade a false error to an mere info as it is expected to happen sometimes and will succeed anyway, so it's not an error


brenzi:
wrong target

PR #1148, opened 2026-06-29
feat: timezone setting in the profile editor (WEB-155)

Claus36868:
## Summary

Implements **WEB-155 — Timezone setting in the profile editor**. Users can now set a `timezone` text record via the profile editor and see it on the public profile header, next to location, as a viewer-relative, DST-aware UTC offset (e.g. `Europe/London (+02:00)` = how far ahead/behind the profile's zone is from the viewer's).

Design reused from the closed PR #568, updated to current code and using **native `Intl`** (no new date dependency — luxon was unnecessary).

## What changed

**Display**
- `timezone` added to `supportedGeneralRecordKeys` — so it lands in the **General** group, categorises as general (not "other"), and flows through the existing `setRecords` multicall unchanged.
- New pure util `utils/getTimezoneOffset.ts` (native `Intl`, DST-aware) — computes the offset between the profile's zone and the viewer's zone at render time; returns `null` for unset/invalid zones.
- `ProfileSnippet.tsx` renders the timezone beside location with a clock icon: `<zone> (±HH:MM)`. Unset/invalid renders nothing.

**Editor**
- New `constants/timezoneOptions.ts` built from `Intl.supportedValuesOf('timeZone')`.
- New `TimezonePicker.tsx` — a searchable IANA picker (thorin `Select` autocomplete) wired through a react-hook-form `Controller`. The picker is rendered wherever a `timezone` General record can be added — the profile editor **and** the registration profile step, create-subname, and reclaim flows (all share the record editor) — so it's never a free-text box.
- New `validators/validateTimezone.ts`, wired into `validatorForRecord`; i18n label + placeholder under `groups.general` in `register.json`.
- Graceful with pre-existing data: a `timezone` record that isn't a canonical IANA id (e.g. an informal `PST` set before this feature, or a zone the runtime's `Intl` build doesn't list) is surfaced in the picker and never blocks saving other edits.

**Untouched (by design):** `getSocialData`, `supportedSocialRecordKeys`, the transaction layer, and `package.json`.

## Tests
- Unit: `getTimezoneOffset.test.ts` (pins the viewer zone; covers positive/negative/same-zone, quarter-hour `+05:45`/`+03:15`, and DST via Jan vs July London), `validateTimezone.test.ts`, and a `categoriseProfileTextRecords` case proving timezone lands in `general` not `other`. `ProfileSnippet.test.tsx` covers render + graceful unset/invalid.
- e2e (`profileEditor.spec.ts`): add → pick from the searchable picker → save → assert the written record and the viewer-relative header offset → clear → assert deletion. Uses a pinned `timezoneId` (viewer `Asia/Kolkata` +05:30, picks `Asia/Tokyo` +09:00 → deterministic, DST-free `+03:30`).

## Decisions on the ticket's open questions
- **Same-zone display:** shows `(+00:00)` (consistent with the `±HH:MM` format) rather than a "same as you" string.
- **Header vs records list:** header-only for the viewer-relative offset; the raw record is still editable in the records list like any general record.
- **AdvancedEditor:** keeps `timezone` as free text (unchanged), per the ticket notes.
- **e2e location:** the editor add/save/display flow is exercised in `profileEditor.spec.ts` (the stateless editor spec with `makeName`), the only place that can drive add → save → public display end-to-end.

## Notes
- `timezone` is an app-level convention, not a standardized ENSIP-5 key.
- `Intl.supportedValuesOf` is ES2022 (fine under the repo's `esnext` lib); `timezoneOptions.ts` degrades to an empty picker on older runtimes rather than throwing.

Ref: WEB-155

https://claude.ai/code/session_01KDLdHpvQQRizSgichgGFKS


sugh01:
## Manual QA report — Timezone in Profile Editor (Claude in Chrome)

**Env:** `localhost:3000`, localhost testnet, wallet `0xf39…92266`. **Viewer TZ:** `Europe/Stockholm` (CEST/+02:00 on 2026-07-07). Names: `other-controller.eth`, `other-registrant.eth`.

**Nav note:** the live profile route is `/<name>` (e.g. `/other-controller.eth`), not `/profile/<name>` as the plan assumed — adjusted accordingly.

### Results — 13/14 PASS, 1 FAIL

| TC | Result | Notes |
|----|--------|-------|
| TC1 Add via picker | ✅ PASS | Timezone under **General**; open-then-search worked; `Asia/Tokyo` selected; tx completed |
| TC2 Record written | ✅ PASS | Records tab: `timezone: Asia/Tokyo` exactly |
| TC3 Header offset | ✅ PASS | `Asia/Tokyo (+07:00)` = Tokyo +09:00 − Stockholm +02:00 |
| TC4 Viewer-relative | ✅ PASS | Simulated viewer `America/New_York` → `(+13:00)`, stored value unchanged |
| TC5 Same-zone | ✅ PASS | Viewer `Asia/Tokyo` → `(+00:00)` |
| TC6 Quarter-hour | ✅ PASS* | Offset padding correct (`Asia/Katmandu (+05:45)`). **Caveat below.** |
| TC7 DST | ✅ PASS | `Europe/London (+01:00)` (BST) for today |
| TC8 Search UX | ✅ PASS | `New York` → label `America/New York`, round-trips to `America/New_York`; gibberish → "No results", no crash |
| TC9 Clear/delete | ✅ PASS | `delete: timezone` tx; record count → 0; header element gone |
| TC10 Unset | ✅ PASS | No clock/text |
| **TC11 Invalid value** | ❌ **FAIL (part 1)** | See below. Part 2 (graceful editing) passed. |
| TC12 Layout | ✅ PASS (desktop) | Location + clock + url on one row, no overlap. Mobile viewport not verified (tooling limitation) |
| TC13 Other flows | ✅ PASS (2/3) | Picker confirmed in **registration** and **create-subname**; **reclaim** not reached |
| TC14 Cancel | ✅ PASS | Unsaved edit + Cancel → no write |

### ❌ TC11 — invalid timezone renders an offset instead of nothing

Set a raw `timezone: PST` record via the Advanced editor. The spec says an invalid/unresolvable zone should render **nothing**; instead the header rendered **`PST (-09:00)`**.

**Root cause (in this PR):** [`isValidTimezone`](src/utils/getTimezoneOffset.ts) treats "anything `Intl.DateTimeFormat` doesn't throw on" as valid. V8's ICU data resolves legacy aliases like `PST`, `EST`, `GMT+5` etc., so they slip through validation and render a bogus label. The same permissive check backs [`validateTimezone`](src/validators/validateTimezone.ts), so these values also aren't rejected on the free-text/programmatic path.

**Suggested fix:** validate against the canonical list the picker already uses (`Intl.supportedValuesOf('timeZone')` in [`timezoneOptions.ts`](src/constants/timezoneOptions.ts)) rather than "didn't throw". Render `null` for anything not in that set. (Keep the existing try/catch as a secondary guard.) Note this must degrade gracefully where `supportedValuesOf` is unavailable — falling back to the current permissive check is acceptable there.

### ⚠️ TC6 caveat — canonical modern zone names missing from the picker

Searching `Kathmandu` / `Kolkata` returns **no results**; only the legacy aliases `Asia/Katmandu` and `Asia/Calcutta` are searchable. The picker is populated from `Intl.supportedValuesOf('timeZone')`, whose output depends on the runtime's ICU build — on this V8 it surfaces the deprecated spellings. Worth deciding whether to (a) accept runtime-dependent naming, or (b) normalize/alias common modern names so users searching `Kolkata`/`Kathmandu` find them. Not a blocker, but a discoverability gap.

### Not verified
- **Mobile/narrow viewport** for TC12 (resize didn't take effect — tester tooling limitation, not a confirmed app issue).
- **Reclaim** flow picker (TC13).
- TC4/TC5 used a JS-level `Intl` override rather than DevTools Sensors (Sensors panel not reachable via the tester's tools) — functionally equivalent for this logic.

src/utils/getTimezoneOffset.ts
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Whether `timeZone` is a time zone that `Intl` can resolve. Unknown zones make
+ * the `Intl.DateTimeFormat` constructor throw a `RangeError`; we treat anything
+ * that throws as invalid so callers can render nothing rather than crash.
+ */
+export const isValidTimezone = (timeZone?: string): timeZone is string => {
+  if (!timeZone || typeof timeZone !== 'string') return false
+  try {
+    // Throws RangeError for unknown time zones.
+    Intl.DateTimeFormat(undefined, { timeZone })
```
sugh01: **QA finding (TC11, FAIL):** this treats "anything `Intl.DateTimeFormat` doesn't throw on" as a valid zone. V8's ICU resolves legacy aliases (`PST`, `EST`, `GMT+5`…), so an informal `timezone: PST` record slips through and the header renders `PST (-09:00)` instead of nothing.

Since this function also backs `validateTimezone`, these values aren't rejected on the free-text/programmatic path either.

Consider validating against the canonical set the picker already builds from `Intl.supportedValuesOf('timeZone')` (see `timezoneOptions.ts`) and returning `false` for anything not in it — keeping this try/catch as a secondary guard, and falling back to the permissive check only where `supportedValuesOf` is unavailable.

sugh01 (approved):

PR #1149, opened 2026-06-30
feat(send-name): explain disabled same-address rows in role search (WEB-102)

Claus36868:
## Summary

Send Name + Edit Roles per-role search already **disabled** a result whose address
already holds the role being set (selecting it would be a no-op), but gave no reason —
the row was just greyed out and showed a generic role `Tag` displaying the address's
*first* role (`userRoles[0]`), which could name an unrelated role and mislead.

This adds a visible, localized message to the disabled row that names **the role that
caused the disable** (`excludeRole`), not the address's first role.

## Changes

- **`SearchViewResult.tsx`** — when `markers.hasRole`, render an `AlreadySetMessage`
  naming `excludeRole` (interpolated from `common:roles.<role>.title`) with a stable
  `data-testid="search-result-already-set-<address>"`. The disabled `<button>` also
  carries the message as `aria-label`/`title` for accessibility. Non-conflicting rows
  are unchanged: enabled, selectable, and they keep their existing role `Tag`.
  `disabled={markers.hasRole}` and the role-match logic are unchanged.
- **`public/locales/en/transactionFlow.json`** — new
  `input.sendName.views.search.alreadySet` = `"This address is already the {{value}}"`
  (under the existing `sendName.search` block, which Edit Roles reuses). Uses the
  `{{value}}` interpolation convention already used in this file (`costValue`,
  `duplicateKey`).

## Scope (deliberate)

Message only — the disable logic is **not** broadened. A v1 unwrapped `.eth` 2LD has
independent handles (registrant/"owner", controller/"manager", ETH address record) that
transfer independently, so disabling on "any role matches" would wrongly block valid
partial transfers. Disable stays an exact, case-insensitive single-role match.

## Tests (Vitest, TDD)

- **`SearchViewResult.i18n.test.tsx`** (new): renders against the **real** locale
  resources (`en/transactionFlow.json` + `en/common.json`) and asserts the actual
  user-visible copy — "This address is already the owner / manager / ETH record",
  including the multi-role case. This verifies AC5 and the code↔JSON
  placeholder/namespace binding that a key-echo i18n mock cannot: a deleted/renamed key,
  a renamed placeholder, or a wrong namespace would change the rendered text and fail.
- **`SearchViewResult.test.tsx`** (new, behaviour/structure): disabled row names
  `excludeRole`; a multi-role address (owner + manager) editing the manager role names
  *manager* and **not** the first role (owner) — guards the badge/disable mismatch (AC3);
  an address holding a *different* role stays enabled with its existing `Tag` and no
  message (AC4); the reason is exposed via `aria-describedby` **without** overriding the
  row's accessible name (AC6); the role match is **exact** (a `parent-owner` holder is not
  disabled when sending `owner`); and AC7 edges — the address is matched
  **case-insensitively on both sides**, and an **undefined-address** record is ignored
  without disabling or crashing. Oracles assert the full explanatory copy, not just the
  role-title fragment, so dropping the message would fail.
- **`SendName.test.tsx`** / **`EditRoles.test.tsx`**: disabled owner / manager / eth-record
  rows show the message naming that specific role; a fresh address stays selectable with no
  message.

Local: `pnpm test` green (1788 passed), `pnpm lint` green, `pnpm lint:types` introduces no
new errors vs the branch baseline (pre-existing errors are confined to unrelated
`*.test`/`e2e` specs that CI does not typecheck).

Ref: WEB-102

### DIFF
--- a/public/locales/en/transactionFlow.json
+++ b/public/locales/en/transactionFlow.json
@@ -363,6 +363,7 @@
         },
         "search": {
           "placeholder": "ENS name or Ethereum address",
+          "alreadySet": "This address is already the {{value}}",
           "views": {
             "error": {
               "message": "Something went wrong. Please try again."
--- a/src/transaction-flow/input/EditRoles/EditRoles.test.tsx
+++ b/src/transaction-flow/input/EditRoles/EditRoles.test.tsx
@@ -205,7 +205,7 @@ describe('EditRoles', () => {
     await waitFor(() => {
       const message = screen.getByTestId('search-result-already-set-0xowner')
       expect(message).toBeVisible()
-      expect(message).toHaveTextContent('roles.owner.title')
+      expect(message).toHaveTextContent('input.sendName.views.search.alreadySet.roles.owner.title')
     })
     await userEvent.click(screen.getByRole('button', { name: 'action.cancel' }))
 
@@ -216,7 +216,9 @@ describe('EditRoles', () => {
     await waitFor(() => {
       const message = screen.getByTestId('search-result-already-set-0xmanager')
       expect(message).toBeVisible()
-      expect(message).toHaveTextContent('roles.manager.title')
+      expect(message).toHaveTextContent(
+        'input.sendName.views.search.alreadySet.roles.manager.title',
+      )
     })
     await userEvent.click(screen.getByRole('button', { name: 'action.cancel' }))
 
@@ -227,7 +229,9 @@ describe('EditRoles', () => {
     await waitFor(() => {
       const message = screen.getByTestId('search-result-already-set-0xeth-record')
       expect(message).toBeVisible()
-      expect(message).toHaveTextContent('roles.eth-record.title')
+      expect(message).toHaveTextContent(
+        'input.sendName.views.search.alreadySet.roles.eth-record.title',
+      )
     })
     await userEvent.click(screen.getByRole('button', { name: 'action.cancel' }))
   })
--- a/src/transaction-flow/input/EditRoles/EditRoles.test.tsx
+++ b/src/transaction-flow/input/EditRoles/EditRoles.test.tsx
@@ -195,6 +195,56 @@ describe('EditRoles', () => {
     await userEvent.click(screen.getByRole('button', { name: 'action.cancel' }))
   })
 
+  it('should show an already-set message naming the role being edited', async () => {
+    render(<EditRoles data={{ name: 'test.eth' }} dispatch={mockDispatch} onDismiss={() => {}} />)
+
+    await userEvent.click(
+      within(screen.getByTestId('role-card-owner')).getByTestId('role-card-change-button'),
+    )
+    await userEvent.type(screen.getByTestId('edit-roles-search-input'), 'owner')
+    await waitFor(() => {
+      const message = screen.getByTestId('search-result-already-set-0xowner')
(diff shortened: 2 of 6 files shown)


sugh01:
## ✅ Manual QA — browser-driven, real UI (all pass)

Tested end-to-end through the running app (local `dev:glocal`, anvil chainId 1337, mock wallet as ACCT0 `0xf39F…2266`) against seeded fixtures. Verified the **exact** user-visible strings, disabled/enabled state, the multi-role oracle, a11y wiring, and Send Name / Edit Roles parity.

**Fixtures:** `other-registrant-2.eth` (owner=ACCT0, manager=ACCT1 `0x7099…79C8`, eth-record=ACCT0) and `other-registrant.eth` (all three roles = ACCT0).

| # | Case | Result |
|---|------|--------|
| Pre | Ownership tab shows owner=ACCT0, manager=ACCT1 (distinct) | ✅ |
| TC1 | Edit Roles → **owner**: disabled ACCT0 row, `This address is already the owner`, no Tag, click is a no-op | ✅ |
| TC2 | Edit Roles → **eth-record** (multi-role): disabled self row, `This address is already the ETH record` — **not** "…owner" | ✅ ¹ |
| TC3 | Edit Roles → **manager**: disabled row, `This address is already the manager`; cross-checked on all-roles name → still "manager", not "owner" | ✅ |
| TC4 | Enabled non-conflicting row (ACCT1 in **owner** search): enabled, `Tag` = **manager**, no message/`title`/`aria-describedby`, clicking proceeds | ✅ |
| TC5 | A11y: `aria-describedby` → message span `id`; `title` = message; **no** `aria-label`; `disabled` on `<button>` | ✅ |
| TC6 | Send Name flow parity: disabled ACCT0 row `This address is already the owner`; ACCT1 row enabled w/ manager Tag | ✅ |
| TC7 | Case-insensitive: all-lowercase ACCT0 address still disabled w/ correct message | ✅ |
| TC8 | Empty/no-results: clean "No results" state, no stray message, no crash | ✅ ² |
| TC9 | Cancel no-op: no tx triggered, roles unchanged after cancel | ✅ |

**Exact strings confirmed verbatim:** `This address is already the owner` / `…the manager` / `…the ETH record`. Accessibility wiring correct (describedby → span id, title, no aria-label). Message correctly names the **edited role** (`excludeRole`), not the address's first role, even for multi-role addresses. Send Name and Edit Roles render identically.

**Notes for the team (not blockers, unrelated to this PR's scope):**
- ¹ On `other-registrant-2.eth` the Edit-Roles modal only exposes the **ETH record** field when the connected wallet holds the manager role (ACCT1 there, not ACCT0). The multi-role assertion for eth-record was therefore verified on `other-registrant.eth` (ACCT0 holds owner+manager+eth-record) — same core oracle, message named the edited role and not the first role.
- ² TC8 logged a background `ContractFunctionExecutionError: the contract function "name" returned no data` from the reverse-registry lookup in the no-results search path. No visible UI impact / no wrong message — flagging as possibly noisy/unnecessary network call on failed searches (pre-existing `useSimpleSearch` behavior, not introduced by this PR).

_Two anchor screenshots captured: disabled owner row (TC1) and enabled row with "manager" Tag (TC4)._

🤖 QA executed via Claude in Chrome

sugh01 (approved):

PR #1150, opened 2026-07-01
feat(avatar): browse NFT avatars from manager, owner and eth-record addresses (WEB-95)

Claus36868:
Ref: WEB-95

## What & why

Users could only source an NFT avatar from the connected wallet or the name's ETH address record — a binary toggle. This lets them browse and pick NFT avatars held by **any** of the name's associated addresses: the connected wallet, the manager (or wrapped owner), the owner/registrant, and the ETH address record.

## Changes

- **`AvatarNFT.tsx`** — replaced the local `useProfileAddresses` helper with a pure `useAvatarSources(name, connectedAddress)` hook that returns an ordered, case-insensitively deduped source list `[connected wallet, ...useOwners rows, eth record]`. The binary toggle button is replaced with a thorin `LegacyDropdown` source selector that sets `selectedAddress`. The connected wallet stays the default (regression-safe) and the selector is hidden when a name has a single source. The fetch layer, pagination, grid, selected-NFT confirm screen, and the `eip155` URI builder are untouched.
- **`transactionFlow.json`** — repurposed the existing `nft.address.owned`/`other` keys to `"Your wallet"` / `"ETH record"`; manager/owner labels come from `useOwners` via the `common` namespace.
- **`AvatarNFT.test.tsx`** — extended the Vitest suite: default source = connected wallet (regression guard), per-source refetch with the selected address as `owner`, case-insensitive dedup, wrapped-name handling (single wrapper owner, no separate registrant), and an end-to-end pick from a non-connected source calling `handleSubmit` with the `eip155` URI.

## Notes / scope

- **No new fetch or service** — `getNfts` already accepts any address; changing `selectedAddress` auto-refetches via the React Query key.
- **No ownership check** (unchanged): external spec-compliant clients (metadata.ens.domains, wallets) verify NFT ownership against the name's associated addresses. The connected wallet still signs the `setRecords` transaction and the `canEditRecords` permission gate is unchanged — browsing another address's NFTs does not change who can write records.
- The `eip155` chain id remains hardcoded to `1` in the URI builder (pre-existing, out of scope).

## Testing

- `pnpm test src/components/@molecules/ProfileEditor/Avatar/AvatarNFT.test.tsx` — 14/14 pass.
- `pnpm lint` → exit 0. `pnpm lint:types` → the change adds zero type errors (repo has pre-existing baseline errors in unrelated `.test`/`.spec` files).

### DIFF
--- a/public/locales/en/transactionFlow.json
+++ b/public/locales/en/transactionFlow.json
@@ -15,8 +15,8 @@
             "loadError": "NFT cannot be loaded",
             "noNFTs": "No NFTs found for this address.",
             "address": {
-              "owned": "Owned",
-              "other": "Other"
+              "owned": "Your wallet",
+              "other": "ETH record"
             },
             "selected": {
               "title": "Selected NFT",
--- a/src/components/@molecules/ProfileEditor/Avatar/AvatarNFT.test.tsx
+++ b/src/components/@molecules/ProfileEditor/Avatar/AvatarNFT.test.tsx
@@ -1,11 +1,20 @@
 /* eslint-disable @typescript-eslint/naming-convention */
-import { fireEvent, mockFunction, render, screen, userEvent, waitFor } from '@app/test-utils'
+import {
+  fireEvent,
+  mockFunction,
+  render,
+  screen,
+  userEvent,
+  waitFor,
+  within,
+} from '@app/test-utils'
 
 import * as ReactQuery from '@tanstack/react-query'
 import React from 'react'
 import { beforeEach, describe, expect, it, Mock, vi } from 'vitest'
 import { useAccount, useClient } from 'wagmi'
 
+import { useNameDetails } from '@app/hooks/useNameDetails'
 import * as UseInfiniteQuery from '@app/utils/query/useInfiniteQuery'
 
 import { makeMockIntersectionObserver } from '../../../../../test/mock/makeMockIntersectionObserver'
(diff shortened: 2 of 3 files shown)


sugh01:
## Manual browser QA — avatar NFT source selector ✅ 6/6 PASS

Tested end-to-end through the real UI on a local `dev:glocal` env (anvil chainId 1337, mock wallet auto-connected as deployer `0xf39F…2266`), against fixtures verified on-chain via the indexer:

- **`other-controller.eth`** — manager=ACCT0 (editable), registrant/eth-record=ACCT1 → sources dedup to **["Your wallet" ACCT0, "Owner" ACCT1]** → selector **shows**.
- **`other-registrant.eth`** — all roles = ACCT0 → **1 source → selector hidden**.

| # | Scenario | Result |
|---|----------|--------|
| TC1 | Multi-source name shows selector; default label **"Your wallet"**; exactly **2** options ("Your wallet", "Owner"), no dup Manager/ETH-record rows; first `getNfts` uses `owner=0xf39F…2266` (connected wallet) | ✅ PASS |
| TC2 | Selecting **"Owner"** updates the button label and fires a new `getNfts` with `owner=0x7099…79C8` (ACCT1) | ✅ PASS |
| TC3 | Single-source name (`other-registrant.eth`) renders the search input with **no** selector; sole request uses ACCT0 | ✅ PASS |
| TC4 | Dedup cross-check: ACCT0 appears once (as "Your wallet", not also "Manager"), ACCT1 once (as "Owner", not also "ETH record") | ✅ PASS |
| TC5 | Search input works alongside the selector; source change still refetches with an active query | ✅ PASS |
| TC6 | Cancel is a no-op — no NFT staged, Save stays disabled, records unchanged on reopen | ✅ PASS |

No crashes, no console errors related to the selector, no duplicate/missing options.

### Note on the `/v1/mainnet/` requests (QA env workaround, not a behavior change)
The `getNfts` worker (`ens-nft-worker`) doesn't serve the local `localhost` network — it returns HTTP 530, which leaves the picker stuck on its "Loading your NFTs" spinner (pre-existing gating: the search input + selector only render after a successful fetch; **not changed by this PR**). To exercise the UI, QA used a **temporary local-only override forcing the worker network to `mainnet`** (not committed) so the fetch resolved. Handy side effect: the two anvil addresses hold distinct real mainnet NFTs (`BeWater Genesis Developer Tier 1` for ACCT0, `Test Token` for ACCT1), making the per-source refetch **visually** verifiable, not just via the request param. This override affected only which grid contents loaded — the feature under test (selector presence, labels, dedup, default, refetch) behaved correctly independent of it.

Screenshots captured during the run: (1) open source dropdown on `other-controller.eth` showing exactly the two options with "Your wallet" as default; (2) `other-registrant.eth` NFT view with the search input and no selector.

sugh01 (approved):

PR #1152, opened 2026-07-01
feat: cap subname extend at parent expiry with Max affordance (WEB-110)

Claus36868:
## What & why

Ref: WEB-110

A wrapped subname with `PARENT_CANNOT_CONTROL` burned can only be extended **up to its parent's expiry**. On-chain, `NameWrapper.extendExpiry` **silently clamps** the requested expiry down to the parent's wrapper expiry — it does not revert. The extend date-picker previously had only a lower bound, so for a subname a user could pick a date **beyond** the parent's expiry: the UI promised a longer extension than the chain would grant.

This caps the subname extend picker at the parent's expiry and adds a one-tap **Max**, so the duration / new-expiry shown in the UI always equals the value actually written on-chain.

## Changes

- **`Calendar.tsx`** — optional `max?: number` mirroring `min`: native `max` attribute + clamp the `onChange` value down to `max`. `maxDate` is derived live (not `useState`-captured) so an async-resolved cap isn't permanently disabled. No-op when `max` is undefined.
- **`DateSelection.tsx`** — optional `maxSeconds?: number`: threaded to `Calendar` `max`, caps the years counter (`maxValue={maxYears}`) and the years-normalisation, adds a clamp effect for an async-resolved cap, hides the years toggle when headroom < 1 year, and adds a **Max** link (`data-testid="date-selection-max"`) that sets the extension to exactly `maxSeconds`.
- **`ExtendNames-flow.tsx`** — reads the parent's **wrapper** expiry via `useParentBasicName().wrapperData.expiry.value` (for a `.eth` 2LD parent this already includes the +90d grace — the exact on-chain clamp ceiling; plain `useExpiry` would be 90 days short). Computes `maxSeconds` for the subname-extension branch only, disables with an "already at maximum expiry" message when headroom < `minSeconds` (avoids the min-bump/max-clamp render loop), and clamps the dispatched `expiryTimestamp` to the parent expiry as a backstop.
- **i18n** — adds `calendar.max` (common) and `input.extendNames.maxExpiryReached` (transactionFlow) across en/es/pt/ru/uk/zh.

`.eth` 2LD extension, registration, and multi-name/bulk extension are unchanged (no cap): when `maxSeconds` is undefined the new code paths reduce exactly to prior behaviour.

## Tests

- `Calendar`: renders/omits `max`, clamps a picked date down to `max`.
- `DateSelection`: Max link sets `seconds = maxSeconds` exactly; clamp pulls a too-large selection down; years toggle hidden when headroom < 1 year; no Max link when uncapped.
- `ExtendNames-flow`: Max on a subname dispatches `expiryTimestamp` == parent expiry (non-year-aligned parent); "already at max expiry" disables + shows the message; no Max for 2LD.

Local: `pnpm test` green (1794 passed), `pnpm lint` clean, no new `tsc` errors introduced by this change.

## ⚠️ Dependency / merge order

This PR is **stacked on top of PR #1126 "pcc subname expiry extension support"** (base branch `pcc-subname-extend-support`), which adds the whole subname-extend path (`extendSubnameExpiry`, `canExtendPccSubname`, the `isSubnameExtension` branch, `useParentBasicName`). On `main` today subnames can't be extended at all, so there is nothing to cap without it.

**Merge order:** land #1126 first, then **rebase this branch onto `main`** and merge as an isolated follow-up. Keeping the base as `pcc-subname-extend-support` scopes this diff to just the cap/Max work.

Note: some pre-existing `tsc`/e2e-infra issues on the `pcc-subname-extend-support` base (unrelated to this change) are inherited here; they are #1126's / a separate follow-up's concern.

## Out of scope (per ticket)

- The Sepolia `isRenewalDisabled` interaction (PCC subname extend incorrectly blocked when the ETHRegistrarController is removed post-v2-beta) — a known follow-up noted in the ticket; mainnet unaffected, fails safe.


sugh01:
## QA: end-to-end browser test — all scenarios pass ✅

Tested this branch through the real UI against a local chain (anvil, chain 1337 + ponder indexer), driving the app in a browser with a real wallet. **9/9 scenarios pass.** No defects found in this PR.

### Environment
- Local dev server, account `0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266`, chain date 2026-07-13, TZ Europe/Stockholm.
- `BaseRegistrar.controllers(ETHRegistrarController) == true` locally, so the `isRenewalDisabled` short-circuit was not triggered and did not confound results.
- The seeded PCC subnames were all unusable as subjects — **every one already sits exactly at its parent's expiry** (headroom 0), i.e. the on-chain clamp this PR is about, already applied by the seed. I minted fresh subnames on-chain to cover the branches:

| Subject | Expiry | Parent cap | Headroom | `maxYears` |
|---|---|---|---|---|
| `qa-3y.emancipated-to-delete.eth` | 2027-06-02 | 2030-06-01 | 1095 d | 2 |
| `qa-200d.wrapped-expired-subnames.eth` | 2026-11-14 | 2027-06-02 | 200 d | 0 |
| `qa-max.wrapped-expired-subnames.eth` | 2027-06-02 | 2027-06-02 | 0 | — |
| `other-registrant.eth` (2LD control) | 2027-03-04 | none | — | — |

### Results

| # | Scenario | Result |
|---|---|---|
| S0 | 2LD regression: no Max link, uncapped years (set to 5), rent price shown, extension transacted | PASS |
| S1 | Default on open: years view, counter 1, new expiry June 2 2028 | PASS |
| S2 | Years counter capped at 2; `+` disabled, value does not exceed cap | PASS |
| S3 | **Max** → date view, 2030-06-01, "2 years, 11 months extension." | PASS |
| S4 | Typed `2035-01-01` into the date field → clamped back to 2030-06-01 | PASS |
| S5 | 200 d headroom → years toggle correctly **absent**, date-picker only | PASS |
| S6 | **Max** → 2027-06-02; typed `2029-01-01` → clamped back to 2027-06-02 | PASS |
| S7 | 0 headroom → "This subname is already at its maximum expiry.", confirm disabled | PASS |
| S8 | **Max → confirm → execute** on a real transaction | PASS |

### The important one (S8), verified on-chain rather than in the UI

After a **Max** extend, I read `NameWrapper.getData` directly:

```
qa-tx.emancipated-to-delete.eth  expiry: 1811899182 -> 1906507182
parent (emancipated-to-delete.eth) expiry: 1906507182
match: true   overshoot: false
```

The submitted `expiryTimestamp` equals the parent's wrapper expiry **exactly, to the second**, with no overshoot. The duration/new-expiry the UI promises is bit-for-bit what the chain writes — which is precisely the bug this PR set out to fix. Using `wrapperData.expiry` (rather than `useExpiry`) is doing its job: the +90 d grace is included, so the cap is the true on-chain clamp ceiling.

### Two notes (neither is a bug in this PR)

**1. The years counter structurally cannot reach the cap.** With 1095 days of headroom, `maxYears` computes to **2**, not 3 — it's a calendar diff anchored at the subname's expiry, and 2027-06-02 → 2030-06-01 is one day short of three calendar years. So the stepper tops out ~1 year below the true ceiling and **"Max" is the only way to consume the full headroom**. The UI is never *wrong* (every displayed value matches what would be written), and this follows from the code as written — just flagging that Max isn't a convenience here, it's load-bearing. Intended?

**2. Missing i18n key — belongs to #1126, not this PR.** The success toast for a subname extension renders a raw key:

> Your `'transaction.description.extendSubnameExpiry'` transaction was successful

`transaction.description.*` lives in `public/locales/en/common.json`, which has `extendNames` but no `extendSubnameExpiry`. The key is absent on `main` **and** on the base branch `pcc-subname-extend-support`, and this PR's only `extendSubnameExpiry` references are code, never locale — so it's inherited from #1126 (which added `extendSubnameExpiry.ts`). One-line fix in `common.json`, over on that PR.

### Out of scope, confirmed correct
2LD extension is genuinely unchanged (no cap, no Max link, rent price still shown, years uncapped), so the subname-only scoping does not leak. No "renewal disabled" banner appeared anywhere.

<sub>Pre-existing console noise unrelated to this PR: `useReverseRegistryName` throws `ContractFunctionExecutionError: the contract function "name" returned no data` on every page (the test account has no reverse record). Reproduces on surfaces this PR doesn't touch.</sub>

sugh01 (approved):

PR #1153, opened 2026-07-16
feat: implement removeStaleWrappedNameData utility

f1lander:
## Summary
- Fixes WEB-613: on address / My Names, some names showed a greyed-out Owner tag and a missing Manager tag (e.g. `ucles.eth`, `theblackparade.eth`).
- Root cause: the ENS subgraph can keep stale wrap data (`fuses` / `wrappedOwner`) after a name is no longer actually wrapped (e.g. expired wrapped name re-registered unwrapped). “Stale” = leftover indexer data that no longer matches on-chain state.
- The UI treats any wrap data as “this name is wrapped,” so it only shows Owner based on `wrappedOwner`. If you’re the real registrant/manager but not that stale wrapped owner, Owner looks grey and Manager disappears — even though on-chain you own/manage the name.
- Fix: before rendering names from `useNamesForAddress`, strip wrap fields when the registry owner is not the NameWrapper contract (a name is only truly wrapped if NameWrapper owns it). Then normal Manager/Owner tags render from registry roles.

## Test plan
- [ ] Local mainnet: open http://localhost:3000/0x341476B838889975cF3D204dfc8dda876932Ef99 — `ucles.eth` shows blue Manager + Owner
- [ ] Local mainnet: open http://localhost:3000/0xf9e21edb4dfc9ff648da6ded3864b79e626e578a?search=theblackparade — blue Manager + Owner (compare with prod grey Owner)
- [ ] Genuinely wrapped names still show a single correct Owner/Manager tag for the wrapped owner
- [ ] `pnpm exec vitest run src/utils/removeStaleWrappedNameData.test.ts`

### screenshots

image

image

### DIFF
--- a/e2e/specs/stateful/un-normalised-name.spec.ts
+++ b/e2e/specs/stateful/un-normalised-name.spec.ts
@@ -4,7 +4,7 @@ import { test } from '../../../playwright/index.js'
 
 test.describe('Un-normalised Name Display', () => {
   const address1 = '0x0c54FcCd2e384b4BB6f2E405Bf5Cbc15a017AaFb'
-  const address2 = '0x43e47385f6b3f8bdbe02c210bf5c74b6c34ff441'
+  // const address2 = '0x43e47385f6b3f8bdbe02c210bf5c74b6c34ff441'
 
   test(`should check correct name is showing for correct address @mainnet`, async ({ page }) => {
     // Enter ENS app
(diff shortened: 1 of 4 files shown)


laurgk (approved):
Good on testing - just waiting for SG to double check code

sugh01 (approved):

PR #1160, opened 2026-07-30
Add Claude-driven manual test workflow for PRs

sugh01:
Runs an exploratory manual test on every PR: boots the local ENS stack, drives the real app in a browser via the Playwright MCP server, and posts a report as a PR comment.

An MCP-driven browser has none of the Playwright fixtures the e2e suite relies on, so init-page.cjs restores the two that matter via MCP's --init-page hook: it injects the same mock wallet the e2e suite uses (@ensdomains/headless-web3-provider, every request kind pre-authorized so nothing needs manual approval) and installs the page clock at the anvil block timestamp, since local chain time is not real time and every name would otherwise look expired. It also exposes a control server on :8546 for switching accounts and travelling in time from the shell.

Supporting pieces:
- wait-for-stack.mjs gates on contracts actually having code on chain and the subgraph reaching the chain head. A leftover .env.local is not proof the deploy finished.
- smoke.mjs boots a real MCP server, loads the app and asserts the wallet, clock and control server work, before any Claude usage is spent. It spawns from the same MCP config the session gets, so the two cannot drift.
- pnpm seed seeds fixtures through the existing makeName helper, without launching a browser.
- The run book always exercises register and extend, so a PR with no user-facing change still gets meaningful coverage.

Long-running processes are started with setsid: each step is its own shell session, and a merely backgrounded child is torn down with it.

### DIFF
--- a/.github/claude/manual-test-instructions.md
+++ b/.github/claude/manual-test-instructions.md
@@ -1,0 +1,173 @@
+# Claude manual-test run book
+
+You are doing exploratory **manual** testing of a pull request against a fully local ENS stack,
+driving a real browser through the Playwright MCP server. You are not writing automated tests.
+
+## Environment (already running — verify, don't rebuild)
+
+| Thing | Where | Notes |
+| --- | --- | --- |
+| App (Next.js dev server) | http://localhost:3000 | started with `pnpm dev:glocal` |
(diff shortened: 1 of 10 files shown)


TwistedCrafts (commented):
Looks good how's it working

